PR #85, opened 2021-12-17
add typedoc

adhusson:


espendk:
@adhusson Is this PR still relevant?

adhusson:
I think you independently added typedoc, right? If yes, then no :)

espendk:
Right ;-)

PR #104, opened 2022-01-06
doc/mgv.js: add examples of using custom eth provider with JS SDK

akettal:


packages/mangrove.js/examples/provider-initialization.mjs
```diff
@@ -0,0 +1,62 @@
+// do not forgot to `yarn add web3@^1.6.1` before launching this script
```
akettal, line 1: FYI, I chose not to include `web3` in the dependencies (even in devdeps) because it's a bit of a stretch to make all users to download it when they probably won't be using this script.

Issue #111, opened 2022-01-08
Connect to a logic without a market / connect to a market without a logic


jkrivine:
Currently using the API on the maker side can only be done by connecting to an offer logic on a specific market using `mgv.MakerConnect({address, base, quote})`. 
However there are use cases where one wants to connect to a logic to read or set its mutable values. There are also use cases where one wants to market make using a simple wallet.
 
Maybe the following could be accessible:
* `const mkr = mgv.connectOfferLogic` to connect to a logic
* `const mkr = mgv.connectEOA` to connect to a wallet
* `const mkrAB = mkr.connectMarket({base:B,quote:A})` to connect the mkr API to a market
this implies that some functions of the mkr API should return an error if not connected to a market

jkrivine:
An alternative is to have EOA makers using `market` functions to post liquidity. This would be consistent since `market` functions allow EOA to pass taker orders already. Then in `market` one would need to connect to a logic in order to post reactive offers. In summary I propose:
* `const market = mgv.market(...);`
* `market.bid(...)` and `market.ask(...)` for EOA makers
* `const logic = market.connectLogic(....)` to connnect to a specific offer logic
* `logic.bid(...)` and `logic.ask(...)` onchain makers

espendk:
I rather like your second proposal.

One idea for avoiding forcing the user to connect to a market just for connecting to the logic could be something like:

```javascript
    const market = mgv.market(...);
    market.bid(...) / market.ask(...); // for EOA makers
    const logic = MyLogic.connect(...); // to connnect to a specific offer logic
    const logicOnMarket = logic.connectMarket(market); // to connect the logic to a specific market
    logicOnMarket.bid(...) / logicOnMarket.ask(...) // onchain makers
```

The type of `logicOnMarket` could be an extension of `logic` with market operations.

jkrivine:
So `MyLogic` would be an independent class?

jkrivine:
would make sense. It would essentially wrap ethers.js calls to functions of https://github.com/mangrovedao/mangrove/blob/master/packages/mangrove-solidity/contracts/Strategies/interfaces/IOfferLogic.sol

espendk:
> So `MyLogic` would be an independent class?

Yes. I thought it already was, since you mentioned "access the offer logic". Also, that pattern could perhaps make it easier to plug in external offer logics ?

jkrivine:
Yes agreed.

jkrivine:
Addressed in commit #146

PR #137, opened 2022-01-15
Feature/offer logic.ts

jkrivine:
Following discussion from issue #111 I implemented the following:
1/ `OfferLogic` can be obtained by doing `new OfferLogic(mgv, contractAddress)`or  from a `Mangrove` object (in which case the `mgv` field of the logic is automatically passed to the constructor).
`OfferLogic` is the class that contains all interactions with a maker contract that do not involve a market (i.e configuration and views).
2/ One can connect `logic:OfferLogic` to a market and obtain a `LiquidityProvider` object that has a `mangrove`, a `market` and a `logic` in its fields. `LiquidityProvider` class implements all functions that involve an offer logic and a market.


packages/mangrove.js/src/offerLogic.ts
```diff
@@ -0,0 +1,193 @@
...
+    } else {
+      this.address = logic.address;
+      this.contract = logic.connect(this.mgv._signer);
+    }
+  }
+  /**
+   * @note Deploys a fresh MangroveOffer contract
+   * @returns The new contract address
+   */
+  static async deploy(mgv: Mangrove, contractName: string): Promise<string> {
```
espendk, line 66: This signature is very broad (accepts any `contractName`) but the method only works for contracts in the mangrove.js typechain dir. Would be good to clarify this limitation and the intended use in the type and/or comment.

packages/mangrove.js/src/offerLogic.ts
```diff
@@ -0,0 +1,193 @@
...
+   * @note Approve Mangrove to spend tokens on the contract's behalf.
+   */
+  approveMangrove(
+    tokenName: string,
+    amount?: Bigish,
+    overrides: ethers.Overrides = {}
+  ): Promise<TransactionResponse> {
+    const _amount =
+      typeof amount === "undefined"
+        ? ethers.BigNumber.from(2).pow(256).sub(1)
```
espendk, line 94: Maybe use `ethers.constants.MaxUint256`

packages/mangrove.js/src/offerLogic.ts
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * @note Approve Mangrove to spend tokens on the contract's behalf.
+   */
+  approveMangrove(
+    tokenName: string,
+    amount?: Bigish,
+    overrides: ethers.Overrides = {}
+  ): Promise<TransactionResponse> {
+    const _amount =
+      typeof amount === "undefined"
```
espendk, line 93: `amount` is declared, so `amount === undefined` is equivalent - and I think it's clearer :-)

packages/mangrove.js/src/offerLogic.ts
```diff
@@ -0,0 +1,193 @@
...
+      await this.mgv._signer.getAddress(),
+      this.mgv.toUnits(amount, 18),
+      overrides
+    );
+  }
+
+  /** Connects the logic to a Market in order to pass market orders. The function returns a LiquidityProvider object */
+  async connectMarket(
+    p:
+      | Market
```
espendk, line 178: The initial `|` with no left-hand side looks weird to me. What's the point of this?

packages/mangrove.js/src/offerLogic.ts
```diff
@@ -0,0 +1,193 @@
+import * as ethers from "ethers";
+import Market from "./market";
+// syntactic sugar
+import { Bigish } from "./types";
+import { typechain } from "./types";
+
+import { LiquidityProvider as LP, Mangrove } from ".";
```
espendk, line 7: Why the alias? To me it's confusing as I rarely read imports (they're a necessary annoyance) and thus I wouldn't immediately realize the connection between `LP` and the class `LiquidityProvider`.

packages/mangrove-solidity/scripts/mumbai/prodTests/OfferProxy/OfferProxyActivateMarkets.js
```diff
@@ -14,25 +14,23 @@ async function main() {
...
   });
   const offerProxy = (await hre.ethers.getContract("OfferProxy")).connect(
     wallet
   );
 
+  const logic = MgvAPI.offerLogic(offerProxy);
+
   for (const tokenName of ["WETH", "USDC", "DAI"]) {
     const token = MgvAPI.token(tokenName);
     const approval = await token.allowance({
```
espendk, line 27: Perhaps the `OfferLogic.mangroveAllowance` could be used instead? I think it would be more concise + showcase use of that API.

packages/mangrove.js/test/integration/simplemaker.integration.test.ts
```diff
@@ -56,40 +62,34 @@ describe("SimpleMaker", () => {
       mgv._provider.pollingInterval = 250;
       await mgv.contract["fund()"]({ value: toWei(10) });
 
-      const mkr_address = await Maker.deploy(mgv, "SimpleMaker");
-      mkr = await mgv.makerConnect({
-        address: mkr_address,
+      const mkr_address = await OfferLogic.deploy(mgv, "SimpleMaker");
+      const logic = mgv.offerLogic(mkr_address);
```
espendk, line 66: Hoisting `logic` to the suite level (as `lp`) would allow simplifications in the tests: `lp.logic` -> `logic`.
This would also demonstrate, that a `LiquidityProvider` is only required when interacting with a market.

`lp` isn't actually used in most of this suite, so perhaps the suite should be split into two: One for `logic` and one for `lp` ?

packages/mangrove.js/test/integration/simplemaker.integration.test.ts
```diff
@@ -30,19 +30,25 @@ describe("SimpleMaker", () => {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-ignore
       mgv._provider.pollingInterval = 250;
-      const mkr_address = await Maker.deploy(mgv, "SimpleMaker");
-      const mkr = await mgv.makerConnect({
-        address: mkr_address,
+      const mkr_address = await OfferLogic.deploy(mgv, "SimpleMaker");
+      const logic = mgv.offerLogic(mkr_address);
+      const lp = await logic.connectMarket({
```
espendk, line 35: This isn't used?

espendk (changes requested):
I think this API is clearer 👍

I've made some suggestions, but nothing major.

jkrivine (commented):
should be OK now

PR #142, opened 2022-01-17
Feature/add logging to mangrove.js

akettal:
Add logging to mangrove.js lib. Console logging is enabled by default, file logging by updating configuration.


espendk:
@akettal I just had a look at the error mesages in cleaning-bot and updategas-bot ("the bots" in the following).

I think the issue is that mangrove.js/src/util/config.ts you set the `NODE_CONFIG_DIR` env variable to a path inside the mangrove.js package. Since the bot tests load mangrove.js before node-config they end up using the `NODE_CONFIG_DIR` set by mangrove.js which is not meaningful.

It makes sense to use the node-config package in mangrove.js, but as it's a library we need to do so in a slightly different manner than in the bots. The documentation describes how it can be done:  https://github.com/lorenwest/node-config/wiki/Sub-Module-Configuration

In addition to fixing the current issue, this approach will also enable us to make it easy for users to configure mangrove.js via configuration files.

espendk:
I just remembered that there's some more relevant stuff about loading config files inside a library package here: https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#loadfileconfigsdirectory-options

This is useful if we want to have the configuration defaults for mangrove.js in a file (which I rather like :-))

akettal:
> I just remembered that there's some more relevant stuff about loading config files inside a library package here: https://github.com/lorenwest/node-config/wiki/Using-Config-Utilities#loadfileconfigsdirectory-options
> 
> This is useful if we want to have the configuration defaults for mangrove.js in a file (which I rather like :-)

Thanks for the advice. mangrove.js is now configurable through a specific namespace, as stated in the README. There also is a default configuration.

packages/commonlib.js/logs/test.log
```diff
@@ -0,0 +1,2 @@
+{"level":"warn","message":"erngpiezjrgpijes"}
```
espendk, line 1: What's the purpose of this file? I can't see that it's referenced anywhere?

packages/commonlib.js/src/logger.ts
```diff
@@ -18,21 +20,23 @@ export interface BetterLogger extends Logger {
 
 export const createLogger = (
   consoleFormatLogger: Format,
-  logLevel: string
+  logLevel: string,
+  additionnalTransports: Transport[] = []
```
espendk, line 24: Minor type (extra 'n'), should be:

```suggestion
  additionalTransports: Transport[] = []
```

packages/commonlib.js/src/logger.ts
```diff
@@ -51,6 +55,14 @@ export const createLogger = (
   return theLogger as BetterLogger;
 };
 
-export { format };
+// This processor must be used when logging large objects, because of Winston memory consumption in that case
+export const logdataProcessor = (data: Object): string => {
```
espendk, line 59: Hmm, this will cause the given data to always be logged as JSON. However, users may want to use other formats in their transports. To allow this, a more flexible approach would be to make a limited depth copy of the `data` object.

Also, a more telling name would be good. Since the purpose is to limit the amount of data logged, perhaps something like `logDataLimiter` ?

packages/mangrove.js/README.md
```diff
@@ -256,7 +256,37 @@ The precision used when dividing is 20 decimal places.
 
 TODO include transaction options (see here)[https://github.com/compound-finance/compound-js#transaction-options]
 
-## Test
+## Package configuration
+
+mangrove.js uses `config` package for its configuration.
```
espendk, line 261:
```suggestion
mangrove.js uses the [node-config](https://github.com/lorenwest/node-config) package for configuration.
```

espendk (changes requested):
Looks good :-) I've suggested a few things - let me know if it makes sense.

We should adapt the bots to pass their `logLevel` configuration to mangrove.js - that would be a good test of how easy/elegant this can be done.

packages/commonlib.js/logs/test.log
akettal: It was mistakenly commited, will remove it.

packages/commonlib.js/src/logger.ts
akettal: Fixed, thanks 👍

packages/commonlib.js/src/logger.ts
akettal: You're right! Fixed.

PR #155, opened 2022-01-21
Empty commit to test automatic deployment of bots on Heroku

espendk:


espendk:
Apparently, empty commits do not trigger CI and therefore no redeploy on Heroku.
I'll try again with a non-empty commit.

PR #159, opened 2022-01-21
Add ability to use any fillWants on buy/sell orders.

adhusson:


packages/mangrove.js/src/types/index.ts
```diff
@@ -33,6 +33,7 @@ export interface MarketParams {
 export type Bigish = Big | number | string;
 
 export type TradeParams = { slippage?: number } & (
-  | { volume: Bigish; price: Bigish }
-  | { wants: Bigish; gives: Bigish }
+  | { volume: Bigish; price: Bigish | null }
```
espendk, line 36: We use `undefined` in other places, but maybe `null` is preferable? We should be consistent, so let's discuss and then make all of mangrove.js consistent :-)

packages/mangrove.js/src/types/index.ts
```diff
@@ -33,6 +33,7 @@ export interface MarketParams {
 export type Bigish = Big | number | string;
 
 export type TradeParams = { slippage?: number } & (
```
espendk, line 35: A bit off-topic, but I'm surprised this type lives here and not in market.ts?

packages/mangrove.js/src/market.ts
```diff
@@ -391,9 +409,25 @@ class Market {
...
-    let _wants =
-      "price" in params ? _gives.mul(params.price) : Big(params.wants);
+    let _wants, _gives, fillWants;
+    if ("price" in params) {
+      if ("volume" in params) {
+        _gives = Big(params.volume);
+        _wants = params.price === null ? 0 : _gives.mul(params.price);
+        fillWants = false;
+      } else {
+        _wants = Big(params.total);
```
espendk, line 419: Just checked and `Big(undefined)` doesn't throw, so we should probably check that `params.total` has been set and throw otherwise.

packages/mangrove.js/src/market.ts
```diff
@@ -391,9 +409,25 @@ class Market {
...
+        _gives =
+          params.price === null
+            ? Big(2).pow(256).minus(1)
+            : _wants.div(params.price);
+        fillWants = true;
+      }
+    } else {
+      _wants = Big(params.wants);
```
espendk, line 427: `wants`  and `gives` should probably also be validated.

packages/mangrove.js/src/market.ts
```diff
@@ -357,9 +358,25 @@ class Market {
    * ```
    */
   buy(params: TradeParams): Promise<Market.OrderResult> {
```
espendk, line 360: See comments for `sell`

espendk (changes requested):
nice improvement :-)

packages/mangrove.js/src/types/index.ts
adhusson: Will leave null for now because otherwise it makes `price` optional right? And then you could easily forget to write the `price` argument and unintentionally run an order that accepts any price.

packages/mangrove.js/src/types/index.ts
adhusson: It seems to be used all over the place (liquidityProvider.ts, mangrove.ts, market.ts, mgvtoken.ts, offerLogic.ts, semibook.ts, index.ts)?

espendk: When I do a whole project search on `TradeParams` I only get this result + 3 in market.ts ?

billede

packages/mangrove.js/src/types/index.ts
espendk: Understood. I was thinking about this being used as a return type, but it clearly isn't.

I think we should still discuss the whole `undefined` vs `null` thing, but let's do that in the team; I'll put it on the agenda for Monday :-)

packages/mangrove.js/src/market.ts
adhusson: I'm missing something, shouldn't the typing ensure params.total has been set?

espendk (approved):

packages/mangrove.js/src/types/index.ts
adhusson: My bad I thought you meant Bigish! Ok moving it.

packages/mangrove.js/src/market.ts
espendk: Right, reading the type I can convince myself that that must be the case. But I'd prefer a more explicit and defensive approach - but that's just an opinion 😅

PR #177, opened 2022-01-27
feat(market): Add options for initially loading offers matching desired volume or price

### LINKED ISSUE #121: `bookOptions` optional parameter of `market()` should allow one to load a market with a volume/price limit
When connecting the API to a market, one does not know a priori how many offers should be loaded. The user knows however how much liquidity she wants to access:
* Taker side
  * If selling `V` base tokens at price `p` one wants to load at least a volume of `V*p` bids.
  * If buying `V` base tokens at price `p` one wants to load at least a volume of `V*p` asks.
* Maker side
  * If posting an offer at price `p` one wants to load at least the offers that are better than `p` (in order to get a meaningful pivot)

Right now one can estimate the number of offers to load using `density`, but this is a coarse approximation.
A very useful feature to have would be to access a market using:

```javascript
MgvAPI.market({base:"WETH", quote:"DAI", volume:1000});
```
or
```javascript
MgvAPI.market({base:"WETH", quote:"DAI", price:0.01});
```

espendk:
Adds new `Market` options as specified in #121:
- `desiredPrice`: allows one to specify a price point of interest. This will cause the cache to initially load all offers with this price or better.
- `desiredVolume`: allows one to specify a volume of interest. This will cause the cache to initially load at least this volume (if available). The option uses the same specification as for `estimateVolume`: `desiredVolume: { given: 1, what: "base", to: "buy" }` will cause the asks semibook to be initialized with a volume of at least 1 base token.

Closes #121.

### DIFF
--- a/packages/mangrove.js/CHANGELOG.md
+++ b/packages/mangrove.js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next version
 
+- New `Market` options:
+  - `desiredPrice`: allows one to specify a price point of interest. This will cause the cache to initially load all offers with this price or better.
+  - `desiredVolume`: allows one to specify a volume of interest. This will cause the cache to initially load at least this volume (if available). The option uses the same specification as for `estimateVolume`: `desiredVolume: { given: 1, what: "base", to: "buy" }` will cause the asks semibook to be initialized with a volume of at least 1 base token.
+
 # 0.1.0 (January 2022)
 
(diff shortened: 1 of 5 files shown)


espendk:
@jkrivine I hope this matches what you had in mind?

I wasn't sure whether this was too advanced to document in README.md - perhaps it should only be documented in Gitbook? Let me know what you think and then I'll add the appropriate documentation.

jkrivine:
This is awesome.

PR #180, opened 2022-01-28
Try to accept both provider&signer objects. Will fail if signer subcl…

adhusson:
…ass cannot `connect`.

### DIFF
--- a/packages/mangrove.js/src/eth.ts
+++ b/packages/mangrove.js/src/eth.ts
@@ -81,8 +81,14 @@ export async function getProviderNetwork(
  * In addition, you can specify
  * - `options.forceReadOnly:boolean` to connect readonly to mangrove. If you don't specify a signer and the provider does not include a signer, you will connect in readonly mode.
  *
- * IMPORTANT if both provider&signer are ethers objects, the signer will connect
- * to the provider (so any signer info on the provider will be ignored).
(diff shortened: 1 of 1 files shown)


packages/mangrove.js/src/eth.ts
```diff
@@ -81,6 +81,9 @@ export async function getProviderNetwork(
  * In addition, you can specify
  * - `options.forceReadOnly:boolean` to connect readonly to mangrove. If you don't specify a signer and the provider does not include a signer, you will connect in readonly mode.
  *
+ * IMPORTANT if both provider&signer are ethers objects, the signer will connect
+ * to the provider (so any signer info on the provider will be ignored).
```
espendk, line 85: We should probably clarify that this only happens if the signer doesn't already have a provider, cf. line 102.

Maybe we should quote ricmoo on how signers and provider are intended to be used in ethers - this might the intentions in the code easier to understand.

> In ethers, there are 3 possible cases:
>
> -   Provider - read-only access
> -   Signer (without a provider) - write-only access
>  -  Signer (with a provider) - read/write access

https://github.com/ethers-io/ethers.js/issues/252#issuecomment-410913959

espendk (changes requested):
lgtm, just added a note and suggestion for the comments.

PR #188, opened 2022-02-06
Feature/mgvjsevents

adhusson:


adhusson:
Note to self: to fix issue of hanging await on `onceWithTxPromise`, remove listener when block containing tx has been processed.

packages/mangrove.js/src/market.ts
```diff
@@ -716,6 +727,44 @@ class Market {
...
+  /* Await until mangrove.js has precessed an event that matches `filter`
+     as part of the transaction generated by `tx`. The goal is to reuse the
+     event processing facilities of market.ts as much as possible but still
+     be tx-specific (and in particular fail if the tx fails).
+     Alternatively one could just use `await (await tx).wait(1)` but then
+     you would not get the context provided by market.ts (current position of a new offer in the OB, for instance).
+  */
+
+  onceWithTxPromise<T>(
+    txPromise: Promise<ethers.ContractTransaction>,
```
espendk, line 739: Isn't there a risk that the tx has been mined at this point? And that the `this.once` will therefore never trigger?

espendk (commented):
Since the listener is still only registered after the tx has been sent, I'm wondering whether it might not have already been mined and the listener therefore never fires?

Jean was able to consistently able to produce the issue. Perhaps his approach could be implemented as an integration test?

packages/mangrove.js/src/market.ts
espendk: @adhusson I've reviewed the latest changes (lgtm), but I don't think they address this?

espendk (changes requested):
I've reviewed the latest changes (lgtm), but I don't think they address the comment about potentially missed events?

We can verify the issue by making an integration test with Jean's approach to producing the problem.

packages/mangrove.js/src/market.ts
espendk: Ok, I think you're right that JS won't switch to any other promises until there's an `await` or the current code terminates.

So for internal stuff we can definitely use this protocol, we just need to be disciplined and ensure there are no `await`s between sending a tx and calling `onceWithTxPromise`.

It's a tricky implicit protocol though, so maybe we shouldn't expose it? Alternatively, we should make the rules for using `onceWithTxPromise` super clear in the docs.

adhusson: Right, that makes sense, I will add info in the comments for now.

PR #195, opened 2022-02-09
Add MgvToken.toFixed method for easy displaying of amounts

adhusson:
Solves #193

### DIFF
--- a/packages/mangrove.js/src/mgvtoken.ts
+++ b/packages/mangrove.js/src/mgvtoken.ts
@@ -71,7 +71,6 @@ class MgvToken {
     if (typeof decimals === "undefined") {
       decimals = this.displayedDecimals;
     }
-    console.log("decimals", decimals);
     return Big(amount).toFixed(decimals);
   }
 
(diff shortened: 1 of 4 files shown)


packages/mangrove.js/src/mgvtoken.ts
```diff
@@ -55,6 +57,24 @@ class MgvToken {
...
+   * ```
+   * token.toFixed("10.123"); // "10.12"
+   * token.toFixed(token.fromUnits("1e7"));
+   * ```
+   */
+  toFixed(amount: Bigish, decimals?: number): string {
+    if (typeof decimals === "undefined") {
+      decimals = this.displayedDecimals;
+    }
+    console.log("decimals", decimals);
```
espendk, line 74: This console.log should be removed, no?

johanleroch:
Looks good to me, as @espendk said, there is one console.log to remove and we can merge 👍

johanleroch (changes requested):
As @espendk already mentionned, let's merge after the console.log() has been deleted 👍

PR #199, opened 2022-02-14
Feature/wmm strat

jkrivine:
v0 of the AMM strat


vd1:
how can I best go about testing the strat by feeding a series of market orders (derived say from a time series of prices)?
do I need to hardhat a local copy of polygon for instance and deploy it on a mangrove market?

PR #202, opened 2022-02-16
Add market convenience estimation method

adhusson:


packages/mangrove.js/src/market.ts
```diff
@@ -57,6 +57,7 @@ namespace Market {
     /** Whether `given` is the market's base or quote. */
     what: "base" | "quote";
   };
+  export type DirectionalVolumeParams = Semibook;
```
johanleroch, line 60: What is it used for?

PR #208, opened 2022-02-21
feat(market): Add `partialFill` flag to `OrderResult`

espendk:
This flag will be true if the order was only partially filled.


espendk:
Will add tests if and when we agree (1) that this is a good idea and (2) think the API is good.

adhusson (approved):
(1) Agreed this is a good idea
(2) API is good
(3) Agree on implementation
:)

PR #221, opened 2022-02-28
fix(mangrove.js/deps): set commonlib.js as a bundledDependency

### LINKED ISSUE #220: Mangrove.js depends on an unpublished packages (`commonlib-js`)
**Describe the bug**

mangrove.js NPM package V0.2.0 has a newly introduced dependency on an unpublished package in the monorepo (commonlib.js), and therefore cannot be installed.

**To Reproduce**

1. Publish a new version of `mangrove.js` NPM package from `c52a56` commit;
2. Add this newly published version as a dependency in a JS/TS project.

**Expected behavior**

`yarn install` in the new project works fine.
(@espendk I supposed the unpublished dependency cannot be resolved at install time, correct me if I'm wrong)

**Additional context**

Bug introduced by the logging feature, which adds a dependency to `commonlib-js` to use a wrapper around Winston.

akettal:
[Bundled dependencies](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundleddependencies) allows to pack an internal and unpublished dependency (commonlib.js) in mangrove.js NPM package.

Fixes #220.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -448,7 +448,7 @@ jobs:
 
   # ==== Job: Build and test mangrove.js ====
   mangrove-js:
-    needs: [security-guard, mangrove-solidity, hardhat-utils]
+    needs: [security-guard, commonlib-js, mangrove-solidity, hardhat-utils]
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


espendk:
I found this nice intro and guide to `bundledDependencies`: https://stackoverflow.com/a/25044361/16614325

.github/workflows/node.js.yml
```diff
@@ -448,7 +448,7 @@ jobs:
 
   # ==== Job: Build and test mangrove.js ====
   mangrove-js:
-    needs: [security-guard, commonlib-js, mangrove-solidity, hardhat-utils]
+    needs: [security-guard, mangrove-solidity, hardhat-utils]
```
espendk, line 451: As discussed we should leave the dependency in, as it is still a dependency even though the package will be bundled.

espendk (changes requested):
See inline comments

.github/workflows/node.js.yml
akettal: 👍

espendk (approved):
lgtm - awesomely simple solution

dontrolle (approved):

akettal:
Closing this MR as it's not fixing the issue, see #220 for more details.

PR #231, opened 2022-03-04
fix(mangrove.js): fix estimateGas underestimation

### LINKED ISSUE #89: `estimateGas` in `market.ts` results in taker out of gas when `takerWants` is too low

dontrolle:
When taker `wants` was low, taker could get an out of gas exception,
when issuing a market order due to `market.estimateGas()`
underestimation (issue #89).

Closes #89.


dontrolle:
I'm marking this PR as Draft for now, as the issue #89 it seeks to address has undergone quite a lot of discussion and ideas for fixes.

Therefore, I'd like one or more reviewers to consider the manner in which this fix tries to address the issue, as well as the implementation details.

In short - this fix seeks to address issue #89 by adding the maximum `gasreq` found in the relevant semibook-cache to the gas-estimate. This is calculated on-demand, when `estimateGas()` is called.

Various other ways (such as caching `maxgasreq`) has been considered, but of the ways this seems the most clean to me now. But please do provide critique of this choice.

In reviewing the PR please consider the following:

1.  By looking at all offers, we potentially open up for a kind of griefing, where somebody posts an extremely high gasreq offer, which is wildly off the mid price (i.e., will never be taken). (Note, this will be griefing against takers who use `mangrove.js`.) A predicate `countOfferForMaxGasPredicate ` has been added in `Semibook.getMaxGasReq()` which allows us to filter out certain offers; but I have not come up with a semantics for this predicate, which I'm happy with. As discussed in #89, it could attempt to disregard offers "to far off the midprice". We can calculate a mid-price - but how much off the mid-price do we allow? Without opening up for other ways of griefing?

2. In the implementation of `Semibook.getMaxGasReq()` we only look at the cache of the semibook. Do we need to guard against the case where:
    -  a take ends up triggering an order *O* outside the OB cache, 
        AND 
    - that order *O* **also** has a gas-estimate that is much larger than every order in the cache?

espendk (commented):
With the disclaimer that I don't fully understand the gas mechanisms, here are two thoughts that crossed my mind when looking at this:

1. Could we perhaps just do a static call (ie simulation) and take the gas use from that?

2. This seems similar to `estimateVolume` where we actually walk the book and simulate a market order. Is it possible to do the same here?

dontrolle:
> With the disclaimer that I don't fully understand the gas mechanisms, here are two thoughts that crossed my mind when looking at this:
> 
>     1. Could we perhaps just do a static call (ie simulation) and take the gas use from that?
Hmmm, that may be an option, actually. I don't actually know how gas usage is calculated with a static call. I'll investigate.  (Or maybe @adhusson or @jkrivine can advise.)
 
>     2. This seems similar to `estimateVolume` where we actually walk the book and simulate a market order. Is it possible to do the same here?

I'll take a look at that.

jkrivine (approved):

jkrivine:
Looks good to me!

adhusson:
It seems the changes look at the actual book cache instead of relying on mangrove's max gas parameter? I'm not clear on why.

dontrolle:
Hey @adhusson - the context of the improvement implemented is here, where @jkrivine and I discussed it: https://github.com/mangrovedao/mangrove/issues/89

As I wrote above in the comment for the PR, we have considered various ideas for improving this, and I have considered various ways to implement this. Including walking the entire book and simulating an order. I'm viewing this as a first attempt at improving the gas-estimates, and I'm entirely fine with discussing further improvements - as I sketch out above. :) This PR had been open for a very long time, though, and the issue was hurting a few tests in the web app, so (Jean and) I decided to get this one merged. 

Let's definetely take a chat to discuss improvement. 👍

Issue #234, opened 2022-03-07
The estimateGas function triggers an error when the density is equal to 0


johanleroch:
**Describe the bug**
`estimateGas` function triggers an error when the density is equal to 0.

**To Reproduce**
1. Launch a fresh local chain with some offers
2. Launch the front on this local chain
3. Create a transaction (buy or sell)
4. An error "division by zero" appears in the console

**Expected behavior**
The tx should go ahead and prompt the metamask confirm panel

**Screenshots**
![telegram-cloud-photo-size-4-5850642268254485030-y](https://user-images.githubusercontent.com/7213352/157093053-b755845d-1f35-4483-940f-4941a9a82764.jpg)
![telegram-cloud-photo-size-4-5850642268254485032-x](https://user-images.githubusercontent.com/7213352/157093382-fd77f8b6-efcc-470a-b398-eba78b6bbb06.jpg)

PR #235, opened 2022-03-07
Fix: The estimateGas function triggers an error when the density is e…

johanleroch:
…qual to 0 #234

### DIFF
--- a/packages/mangrove.js/src/market.ts
+++ b/packages/mangrove.js/src/market.ts
@@ -597,11 +597,14 @@ class Market {
       bs === "buy"
         ? await this.#asksSemibook.getRawConfig()
         : await this.#bidsSemibook.getRawConfig();
-    const maxMarketOrgerGas: BigNumber = BigNumber.from(MAX_MARKET_ORDER_GAS);
+    const maxMarketOrderGas: BigNumber = BigNumber.from(MAX_MARKET_ORDER_GAS);
     const estimation = density.isZero()
-      ? maxMarketOrgerGas
+      ? maxMarketOrderGas
       : offer_gasbase.add(volume.div(density));
-    return estimation.gt(MAX_MARKET_ORDER_GAS) ? maxMarketOrgerGas : estimation;
+
+    if (estimation.lt(maxMarketOrderGas)) return estimation;
+
+    return maxMarketOrderGas;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/mangrove.js/src/market.ts
```diff
@@ -591,18 +591,17 @@ class Market {
...
-    if (estimation.gt(MAX_MARKET_ORDER_GAS)) {
-      return BigNumber.from(MAX_MARKET_ORDER_GAS);
-    } else {
-      return estimation;
-    }
+    const maxMarketOrgerGas: BigNumber = BigNumber.from(MAX_MARKET_ORDER_GAS);
+    const estimation = density.isZero()
+      ? maxMarketOrgerGas
+      : offer_gasbase.add(volume.div(density));
+    return estimation.gt(MAX_MARKET_ORDER_GAS) ? maxMarketOrgerGas : estimation;
```
dontrolle, line 604: Personally, I think something like the following for the last line is a bit more readable - but I'll leave this up to you.

```typescript
...
    if (estimation.lt(maxMarketOrgerGas))
      return estimation;

    return maxMarketOrgerGas;
```

dontrolle (changes requested):
After settling the review-comments below, and the particular note that I think @adhusson or @jkrivine should just nod to the case for `density=0`, I think it's ready to go. :)

As a side-note - the `estimateGas()` func is going to be updated again when PR https://github.com/mangrovedao/mangrove/pull/231 lands. (I'll just rebase, when it's reviewed.)

packages/mangrove.js/src/market.ts
```diff
@@ -591,18 +591,17 @@ class Market {
...
         : await this.#bidsSemibook.getRawConfig();
-    const estimation = rawConfig.local.offer_gasbase.add(
-      volume.div(rawConfig.local.density)
-    );
-    if (estimation.gt(MAX_MARKET_ORDER_GAS)) {
-      return BigNumber.from(MAX_MARKET_ORDER_GAS);
-    } else {
-      return estimation;
-    }
+    const maxMarketOrgerGas: BigNumber = BigNumber.from(MAX_MARKET_ORDER_GAS);
```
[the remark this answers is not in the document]
johanleroch: oups! thanks :)

packages/mangrove.js/src/market.ts
johanleroch: I agree, more readable like this :)

dontrolle (approved):

PR #237, opened 2022-03-09
Feature/test partial fail

johanleroch:
This is a test script allowing me to deploy offers that are partially messed up on the local chain. In order to test the front-end interractions with metamask.
--
### Changes : 

- New deploy script
- Update .env.local.example in order to add context for other devs and make the script more dynamic


espendk (changes requested):
I think this is a great start for setting up local chains!

Could you add a preamble with a description of what the file does and an explanation of its dependency on `TestingSetup`

espendk (approved):
nice :-)

Issue #101, opened 2022-01-05
Add `overrides` parameters in all API functions that produce signed transactions.


jkrivine:
Currenlty many function resulting in a tx do not accept an optional `overrides` argument. As a result one cannot set the `gasLimit` manually.

espendk:
Closely related to another API uniformity issue: #173

PR #240, opened 2022-03-10
Feature/overrides

jkrivine:
This PR addresses issue #101 
It should let users define ethers.js overrides to market orders via the API.
NB: using a user defined `gasLimit` overrides the gasLimit estimates of the API (one assume the user knows what she does)


jkrivine:
Fixing #101

PR #254, opened 2022-03-14
Don't hang if onceWithTxPromise gets a successful tx but no matching …

adhusson:
…event

### DIFF
--- a/packages/mangrove.js/src/market.ts
+++ b/packages/mangrove.js/src/market.ts
@@ -825,17 +825,20 @@ class Market {
   ): Promise<T> {
     return new Promise((ok, ko) => {
       const txHashDeferred = new Deferred<string>();
+      let someMatch = false;
       const _filter = async (
         cbArg: Market.BookSubscriptionCbArgument,
         event: Market.BookSubscriptionEvent,
         ethersEvent: ethers.ethers.providers.Log
       ) => {
-        return (
-          filter(cbArg, event, ethersEvent) &&
-          (await txHashDeferred.promise) === ethersEvent.transactionHash
-        );
+        const match =
+          (await txHashDeferred.promise) === ethersEvent.transactionHash;
+        someMatch = someMatch || match;
+        return match;
       };
-      this.once(cb, _filter).then(ok, ko);
+      this.once(cb, _filter).then(ok, (e) =>
+        ko({ revert: false, exception: e })
+      );
 
       txPromise.then((resp) => {
         // Warning: if the tx nor any with the same nonce is ever mined,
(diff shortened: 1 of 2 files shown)


adhusson:
Note: it would be better with a semaphore, for now we rely on each `filter` awaiting `txHashDeferred.promise;` exactly once.

packages/mangrove.js/src/market.ts
```diff
@@ -821,21 +821,24 @@
...
         ethersEvent: ethers.ethers.providers.Log
       ) => {
-        return (
-          filter(cbArg, event, ethersEvent) &&
-          (await txHashDeferred.promise) === ethersEvent.transactionHash
-        );
+        const match =
+          (await txHashDeferred.promise) === ethersEvent.transactionHash;
+        someMatch = someMatch || match;
+        return match;
```
espendk, line 837: `filter` should be called somewhere here, right?

espendk (changes requested):
See inline comments

espendk:
There's an Issue with the GH Actions workflow that synchronizes assignees and reviewers. I've applied a fix and will use this PR to test it. Sorry for the noise.

packages/mangrove.js/src/market.ts
adhusson: Should be fixed by ea435d125afdc5b4272b59a7ebd158dbbadf71da

adhusson:
> There's an Issue with the GH Actions workflow that synchronizes assignees and reviewers. I've applied a fix and will use this PR to test it. Sorry for the noise.

no worries :)

I should have pushed a fix for your comment. I'll add a proper lock in a later PR.

espendk (approved):

adhusson:
Note on [85c8055](https://github.com/mangrovedao/mangrove/pull/254/commits/85c80556788b075671235d451edcf1c0cea68e5f): ideally each `runExclusive` should be able to run concurrently -- this would require writing a new kind of mutex with two functions `waitForUnlock` and `runExclusiveButDontWait`, with the second function creating a lock if there is not one but not necessitating the mutex to be free to run.

espendk (commented):
Because the mutex only locks the individual `_filter` calls, isn't there a risk that the `afterBlock` code runs before (or in between) the `_filter` code? 

This code is hard to read an reason about and it's even more difficult with the mutex.

Maybe an easier-to-understand approach would be the following:

- Add `beforeEachBlock` and `afterEachBlock` events that can be subcribed to
- In `onceWithTxPromise` do the following:
  1. Register a `beforeEachBlock` handler that locks `mutex`
  2. Register an `afterEachBlock` handler that unlocks `mutex`.
  3. Remove the locking from `_filter` as the before/after block-handlers handle locking


This will also allow parallel processing of `_filter`s.

adhusson:
The `aterBlock` code will always run after all filters for a block have been executed. But yes it's hard to follow.

I don't think the `before`/`after` each block method does what we want. If the mutex is unlocked after all filters have been called, that doesn't imply all filters have "finished" executing. My goal is to detect that the entire promise of each filters is done -- some kind of `Promise.all` across all filters being run for the current block.

espendk (approved):
Much cleaner I think 👍

PR #374, opened 2022-04-22
log inactivity

jkrivine:


espendk (approved):
lgtm - added some block info to the log in the hope that it will allow us to identify issues caused by reorgs

espendk:
Tests timed out, I've restarted them.

PR #382, opened 2022-04-22
wip implementing custom taker

jkrivine:


packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
```diff
@@ -0,0 +1,114 @@
...
+    external
+    payable
+    returns (
+      uint takerGot,
+      uint takerGave,
+      uint bounty,
+      uint offerId
+    )
+  {
+    require(msg.sender != address(this), "CustomTaker/noReentrancy");
```
adhusson, line 57: Not sure what this is trying to do

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
```diff
@@ -0,0 +1,114 @@
...
+      bounty += bounty_;
+      if (takerGot_ == 0) {
+        break;
+      }
+    }
+    if (tko.selling && tko.partialFillNotAllowed) {
+      require(takerGave == tko.gives, "CustomTaker/noPartialFill");
+    }
+    if (tko.selling && tko.partialFillNotAllowed) {
+      require(takerGot == tko.wants, "CustomTaker/noPartialFill");
```
adhusson, line 86: should be >=

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
```diff
@@ -0,0 +1,114 @@
...
+      });
+      takerGot += takerGot_;
+      takerGave += takerGave_;
+      bounty += bounty_;
+      if (takerGot_ == 0) {
+        break;
+      }
+    }
+    if (tko.selling && tko.partialFillNotAllowed) {
+      require(takerGave == tko.gives, "CustomTaker/noPartialFill");
```
adhusson, line 83: conceptually it should be <= no matter `tko.selling` and `tko.partialFillNotAllowed`, which is anyway guaranteed by Mangrove, so that can be removed.

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
```diff
@@ -0,0 +1,114 @@
...
+      require(takerGot == tko.wants, "CustomTaker/noPartialFill");
+    }
+    if (bounty > 0) {
+      (bool noRevert, ) = msg.sender.call{value: bounty}("");
+      require(noRevert, "CustomTaker/transferBountyFailed");
+    }
+    // resting limit order for the residual of the taker order
+    // tko.wants / tko.gives = (tko.wants - totalGot) / new_gives
+    // hence new_gives = (tko.wants - totalGot) * tko.gives / tko.wants
+    if (tko.restingOrder) {
```
adhusson, line 95: should be `if (tko.restingOrder && takerGot < tko.wants)`

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
```diff
@@ -0,0 +1,114 @@
...
+      uint offerId = newOfferInternal({
+        outbound_tkn: inb,
+        inbound_tkn: out,
+        wants: tko.wants - takerGot,
+        gives: ((tko.wants - takerGot) * tko.gives) / tko.wants,
```
adhusson, line 100: Not convinced this is the right approach. Assume the taker accepted a price of 1 and got almost his entire fill for a price of 0.5. You could argue that for the remaining epsilon they want to buy they are ready to pay a price >>> 1. I'm not sure but I think saying `gives: tko.gives - takerGave` is thus also reasonable.

adhusson (commented):
Amazing :) At this point I would include:
* manage native tokens
* tests (in solidity?)

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
jkrivine: Key assumption here is that when going through a resting limit order the ratio `wants/gives` is a strict indication of the price you expect to buy or sell. The semantics being try to fill my order at that price and buy/sell whatever remains (at that same price)

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
jkrivine: The idea here is to revert in case of a partial fill. When selling (`tko.selling`) you passed a marketOrder with `fillWants=false` so you want to revert if `takerGave < tko.gives`

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
jkrivine: see comment below

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
adhusson: OK makes sense

packages/mangrove-solidity/contracts/Strategies/OfferLogics/MultiUsers/CustomTaker.sol
adhusson: Makes sense for taker gives but I would double check that you can't get even 1 Wei above takerwants when fillwants is true.

jkrivine:
> Ok looks good! I would still add tests, manage native tokens (so CustomTaker doubles as a native token proxy for the mangrove frontend) and also maybe use that as an opportunity to move to a new repo and use foundry?

Yes. Will do tests and native token support. I would love some help wrt making a new foundry repo however

PR #390, opened 2022-04-25
feat: Use `ethers.getDefaultProvider` instead of limited protocol-specific providers

espendk:
`getDefaultProvider` provides more flexibility as it detects the protocol and constructs an appropriate provider. Thus by using that instead of `new WebSocketProvider` or `new JsonRpcProvider` the user/caller is free to select the protocol.

### DIFF
--- a/packages/bot-cleaning/src/index.ts
+++ b/packages/bot-cleaning/src/index.ts
@@ -8,7 +8,7 @@ import config from "./util/config";
 import { logger } from "./util/logger";
 
 import Mangrove from "@mangrovedao/mangrove.js";
-import { WebSocketProvider } from "@ethersproject/providers";
+import { getDefaultProvider } from "@ethersproject/providers";
 import { NonceManager } from "@ethersproject/experimental";
 import { Wallet } from "@ethersproject/wallet";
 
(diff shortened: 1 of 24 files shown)


vd1:
usually I do something like 
```
let provider = new ethers.providers.WebSocketProvider(env.parsed.MUMBAI_NODE_KEY);
```
should I do this instead:
```
let provider2 = new ethers.getDefaultProvider(env.parsed.MUMBAI_NODE_KEY)
```
and what are the differences?

espendk:
You should do:

```js
let provider = ethers.getDefaultProvider(env.parsed.MUMBAI_NODE_KEY)
```

The advantage of this (as it says in the description 🙃) that this will detect the protocol specified in the URL and construct an appropriate provider. This your `MUMBAI_NODE_KEY` env var can contain both `http` and `ws` URL's.

vd1:
OK I am using this URL `https://polygon-mumbai.g.alchemy.com/v2/` so I guess the default is an http connector for me? and before I was forcing a `ws` one? And now the choice of connector is in the env file? But which protocol should I prefer?

espendk:
yes, default for that URL would be `JsonRpcProvider` . Lucky that it also worked with WebSockets.
Whether to use http or ws is a tradeoff: Performance seems better with ws, but ethers.js cannot maintain stable ws connections, so for long-running stuff http is the safer choice.

Btw, let's move this discussion to Discord where more people can benefit :-)

vd1:
which channel?

PR #394, opened 2022-04-27
Feat/mgv oracle implement setdensity

dontrolle:
Implement `setDensity` in `MgvOracle`. We ensure that the default density in `MgvOracle` is set s.t. `Mangrove` uses its internal density by default.


adhusson (approved):
Let's go 😃

PR #429, opened 2022-05-05
Feature/advanced taker api

jkrivine:


packages/mangrove.js/src/market.ts
```diff
@@ -442,6 +453,7 @@ class Market {
    */
   buy(
     params: Market.TradeParams,
+    opt: Market.OptTradeParams = {},
```
adhusson, line 456: Looks like the main backward compatibility breaker, or is there another? (same for sell)

packages/mangrove.js/src/market.ts
```diff
@@ -464,13 +476,18 @@ class Market {
...
     }
 
-    const slippage = validateSlippage(params.slippage);
+    const slippage = validateSlippage(opt.slippage);
 
     _gives = _gives.mul(100 + slippage).div(100);
 
     const wants = this.base.toUnits(_wants);
     const gives = this.quote.toUnits(_gives);
-
```
adhusson, line 473: For style points and future-proofing I would add an `else` to that `if` instead of relying on `return`.

packages/mangrove.js/src/market.ts
```diff
@@ -464,13 +476,18 @@ class Market {
...
+    if (opt.restingOrder) {
+      return this.#restingOrder(
+        { gives, wants, orderType: "buy", fillWants, opt },
+        overrides
+      );
+    }
     return this.#marketOrder(
       { gives, wants, orderType: "buy", fillWants },
```
adhusson, line 492: There is something surprising about entirely ignoring `opts` unless `opts.restingOrder` is set. Shouldn't the value be passed to `marketOrder`?

packages/mangrove.js/src/market.ts
```diff
@@ -521,13 +539,18 @@ class Market {
...
+    const slippage = validateSlippage(opt.slippage);
 
     _wants = _wants.mul(100 - slippage).div(100);
 
     const gives = this.base.toUnits(_gives);
     const wants = this.quote.toUnits(_wants);
-
+    if (opt.restingOrder) {
+      return this.#restingOrder(
+        { gives, wants, orderType: "buy", fillWants, opt },
```
adhusson, line 550: Should be `sell`.

packages/mangrove.js/src/market.ts
```diff
@@ -615,6 +638,89 @@ class Market {
...
+      wants: ethers.BigNumber;
+      gives: ethers.BigNumber;
+      orderType: "buy" | "sell";
+      fillWants: boolean;
+      opt: Market.OptTradeParams;
+    },
+    overrides: ethers.Overrides
+  ) {
+    let overrides_: ethers.PayableOverrides;
+    if (opt.provision) {
```
adhusson, line 658: I would throw or warn if value was already in overrides.

packages/mangrove.js/src/market.ts
```diff
@@ -615,6 +638,89 @@ class Market {
...
+    // user defined gasLimit overrides estimates
+    overrides_.gasLimit = overrides_.gasLimit
+      ? overrides_.gasLimit
+      : await this.estimateGas(orderType, wants);
+
+    const response = await this.mgv.orderContract.take(
+      {
+        base: this.base.address,
+        quote: this.quote.address,
+        partialFillNotAllowed: opt.partialFillNotAllowed
```
adhusson, line 672: Since this is a pretty subtle function, I am tending towards forcing every `opt` field to be defined. You really want to make the programmer think about every parameter, no?

packages/mangrove.js/src/market.ts
jkrivine: `value` cannot be in `overrides` by typing constraint (it would be `PayableOverrides`)

packages/mangrove.js/src/market.ts
jkrivine: Not sure because there are very natural default values for those.

packages/mangrove.js/src/market.ts
jkrivine: Yes but since `#marketOrder` don't use any of these options it would be a bit useless.

packages/mangrove.js/src/market.ts
```diff
@@ -644,22 +658,21 @@ class Market {
...
   ) {
-    let overrides_: ethers.PayableOverrides;
-    if (opt.provision) {
-      overrides_ = { ...overrides, value: this.mgv.toUnits(opt.provision, 18) };
-    } else {
-      overrides_ = overrides;
-    }
+    const overrides_ = {
+      ...overrides,
+      value: this.mgv.toUnits(params.provision, 18),
```
adhusson, line 673: What if `params.provision` is not defined here?

jkrivine: provision is no longer optional now

PR #450, opened 2022-05-11
Feat/market subscribe fail returns parsed mgvdata

dontrolle:
`market.subscribe` now returns a user friendly `mgvData` in the case of `OfferFail`.


packages/mangrove.js/test/mocha/config/market-integration-tests.json
```diff
@@ -0,0 +1,11 @@
...
+  "require": [
+    "ts-node/register",
+    "@mangrovedao/hardhat-utils/mocha/hooks/integration-test-root-hooks"
+  ],
+  "recursive": true,
+  "spec": "test/integration/market.integration.test.ts",
+  "timeout": 60000,
+  "reporter": "mocha-multi-reporters",
+  "reporterOptions": "configFile=test/mocha/config/integration-tests-multi-reporters.json"
+}
```
espendk, line 11: This configuration seems to be the same as the one for the other integration test suites (`integration-tests.json`) - how come a separate one is needed?

packages/mangrove.js/test/integration/market.integration.test.ts
```diff
@@ -172,6 +172,49 @@ describe("Market integration tests suite", () => {
...
+    // Note: the "as const" is necessary, else TS insists that 'asks' should be widened as a string, which doesn't match the type - I view this as a failing of the TS type-checker
+    const newFailingOffer = {
+      market,
+      ba: "asks" as const,
+      maker,
+      shouldFail: true,
+    };
+
+    await mgvTestUtil.postNewOffer(newFailingOffer);
+
```
espendk, line 198: Inlining `newFailingOffer` into the `postNewOffer` call would eliminate the need for the cast (and the rant ;-)

We could also introduce a util `postNewFailingOffer` similar to those that already exist for reverting and succeeding offers. Could maybe make the test slightly more concise?

packages/mangrove.js/test/mocha/config/market-integration-tests.json
dontrolle: You're right. Good catch. This was a config, I used temporarily, which was mistakenly commited. I'll remove it.

packages/mangrove.js/test/integration/market.integration.test.ts
dontrolle: Good idea. I'll do that.

Issue #136, opened 2022-01-15
Precision when dividing for `Big.js` module should be set in a centralized fashion


jkrivine:
Currently `Big.js` is imported separately in different modules, but only in `mangrove.ts` we set:
```typescript
import Big from "big.js";
Big.DP = 20; // precision when dividing
Big.RM = Big.roundHalfUp; // round to nearest
```
We should import `Big` from, say `constant.ts` in which the above setting is performed once for all

PR #467, opened 2022-05-13
Only one big.js import

adhusson:
big.js can be configured a single time in mangrove.js.

Issue: if users import parts of mangrove.js directly (but don't import the Mangrove class), the big.js instance will not be properly configured.

edit: addresses #136

### DIFF
--- a/packages/mangrove.js/src/mangrove.ts
+++ b/packages/mangrove.js/src/mangrove.ts
@@ -14,6 +14,10 @@ import { Bigish } from "./types";
 import { LiquidityProvider, OfferLogic, MgvToken, Market } from ".";
 
 import Big from "big.js";
+// Configure big.js global constructor
+Big.DP = 20; // precision when dividing
+Big.RM = Big.roundHalfUp; // round to nearest
+
 import * as ethers from "ethers";
 Big.prototype[Symbol.for("nodejs.util.inspect.custom")] =
   Big.prototype.toString;
--- a/packages/bot-maker-noise/src/OfferMaker.ts
+++ b/packages/bot-maker-noise/src/OfferMaker.ts
@@ -6,8 +6,6 @@ import Big from "big.js";
 import { MakerConfig } from "./MarketConfig";
 import { fetchJson } from "ethers/lib/utils";
 import { clearTimeout, setTimeout } from "timers";
-Big.DP = 20; // precision when dividing
-Big.RM = Big.roundHalfUp; // round to nearest
 
 // FIXME Move to mangrove.js
 export type BA = "bids" | "asks";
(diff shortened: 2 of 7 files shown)


packages/mangrove.js/src/mangrove.ts
```diff
@@ -14,6 +14,10 @@ import { Bigish } from "./types";
 import { LiquidityProvider, OfferLogic, MgvToken, Market } from ".";
 
 import Big from "big.js";
+// Configure big.js global constructor
+Big.DP = 20; // precision when dividing
+Big.RM = Big.roundHalfUp; // round to nearest
+
 import * as ethers from "ethers";
```
espendk, line 21: Maybe it would be cleaner to have initialization code in `index.ts`, i.e. the `main` file pointed to by `package.json` ?

I just tried adding a log statement to `index.ts` and then importing a sub-part of the module. The log statement was executed, so I think it's safe to assume any code in the `main` file is executed regardless of how the module is imported.

```js
// index.ts
console.log("my test");

// node REPL
> var foo = require("./dist/nodejs/mangrove.js");
my test
undefined
```

espendk (changes requested):
see comment in mangrove.ts

packages/mangrove.js/src/mangrove.ts
```diff
@@ -14,6 +14,10 @@ import { Bigish } from "./types";
...
 Big.prototype[Symbol.for("nodejs.util.inspect.custom")] =
   Big.prototype.toString;
```
espendk, line 23: This should probably follow the other BIg initialization code.

adhusson: I don't know what you mean here?

packages/mangrove.js/src/mangrove.ts
espendk: Ok, darn - a bit weird to me, that the module system doesn't have support for module initialization. I guess we could achieve it by following a convention, e.g. all files must include a module initialization file; Would be painful to maintain though...

Regarding placement, ok. It seems weird to me to place initialization in `mangrove.ts` as it's not specific to that class, but actually initializes the whole module. If you think direct imports of sub parts is likely and something we should support, I'd argue we should go for a general solution instead of only supporting that for `mangrove.ts`.

espendk (commented):
see comment

packages/mangrove.js/src/mangrove.ts
adhusson: I will keep it that way for now: it is realistic that someone would use `Mangrove` directly, but other classes are not realistic as they directly or indirectly need a Mangrove instance to be construted. So I think it hits a sweet spot in terms of being safe and not too ugly. 

Regarding generally supporting sub-parts imports, I can't think of a non-painful method yet. Also looking around module side-effects I came across this https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-imports-being-elided-in-my-emit which is... interesting.

BTW another option I tried was to have a Mangrove-only `Big` object. You would `import {Big} from "@mangrovedao/mangrove.js"` and that would be not the same class as `import Big from "big.js"`. It works, and has the advantage of not polluting the user's `Big` with our configuration... but it felt counterintuitive and more likely to lead to confusion.

PR #479, opened 2022-05-23
Move from Winston to Loglevel

adhusson:
Currently replaced Winston with loglevel. If fine, will 

- [ ] publish the commonlib package and 
- [ ] remove mangrove.js' logger.ts

### DIFF
--- a/packages/commonlib.js/package.json
+++ b/packages/commonlib.js/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@mangrovedao/commonlib-js",
+  "name": "@mangrovedao/commonlib.js",
   "version": "0.0.2",
   "author": "Mangrove DAO",
   "description": "Common javascript and typescript library code employed in the Mangrove mono-repo.",
--- a/packages/commonlib.js/package.json
+++ b/packages/commonlib.js/package.json
@@ -25,11 +25,13 @@
     "**/*": "prettier --write --ignore-unknown"
   },
   "dependencies": {
+    "@types/triple-beam": "^1.3.2",
     "config": "^3.3.7",
     "dotenv-flow": "^3.2.0",
     "fast-safe-stringify": "^2.1.1",
     "json-truncate": "^3.0.0",
-    "winston": "^3.7.2"
+    "logform": "^2.4.0",
+    "triple-beam": "^1.3.0"
   },
   "devDependencies": {
     "@espendk/json-file-reporter": "^1.4.2",
--- a/packages/commonlib.js/package.json
+++ b/packages/commonlib.js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mangrovedao/commonlib-js",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "author": "Mangrove DAO",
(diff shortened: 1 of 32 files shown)


packages/mangrove-solidity/config/default.js
```diff
@@ -53,6 +53,11 @@ config.hardhat = {
         enabled: true,
         runs: 20000,
       },
+      outputSelection: {
+        "*": {
+          "*": ["storageLayout"],
+        },
+      },
```
espendk, line 60: This seems unrelated to the rest of the PR?

packages/commonlib.js/package.json
```diff
@@ -25,11 +25,13 @@
     "**/*": "prettier --write --ignore-unknown"
   },
   "dependencies": {
```
espendk, line 27: I'd expect to see `loglevel` added here as it's now a direct dependency.

packages/commonlib.js/src/logger.ts
```diff
@@ -1,65 +1,109 @@
...
-  logLevel: string,
-  additionalTransports: Transport[] = []
-): BetterLogger => {
-  const consoleTransport = new transports.Console({
-    level: logLevel,
-    handleExceptions: true,
-    format: format.combine(
-      format.colorize(),
+  logLevel: string
+  //additionalTransports: Transport[] = []
```
espendk, line 43: Since we're deleting Winston, I think we should remove this - it's no longer relevant.

espendk (changes requested):
Looks good. Have added a few minor suggestions.

packages/mangrove-solidity/config/default.js
```diff
@@ -53,6 +53,11 @@ config.hardhat = {
         enabled: true,
         runs: 20000,
       },
+      outputSelection: {
```
dontrolle, line 56: Apparently my comment from earlier today got deleted, so I'll repost it:
> What does this do?

dontrolle (commented):
Apart from the minor comments, looks good.

packages/commonlib.js/src/logger.ts
adhusson: Do you mean removing the commented out `additionalTransports...` ?

packages/mangrove-solidity/config/default.js
adhusson: This should not be here :)

packages/commonlib.js/src/logger.ts
espendk: Yes, that's what I mean, sorry for being imprecise :-) (in my defense, GitHub is retarded - I've commented on only that line, but it's making it look like I've commented on a two-line diff 🙄)

adhusson:
I have added a fix to properly setup the loglevel, and make mangrove.js use the published version on commonlib.

I think next steps are you give a go to the bots on this branch and if everything is fine we can merge.

espendk (commented):
I've renamed commonlib-js to commonlib.js.

I'll test logging in a bot and then approve.

espendk (approved):
I just checked the cleaning bot and the logs look ok.
We still need to verify once this has been deployed to Heroku.

PR #484, opened 2022-05-30
Feature/simpler logging

adhusson:
refactor: disable mangrove.js logging by default, remove leftover logdataLimiter

### DIFF
--- a/packages/mangrove.js/src/util/logger.ts
+++ b/packages/mangrove.js/src/util/logger.ts
@@ -1,5 +1,10 @@
 import inspect from "object-inspect";
-import { createLogger, CommonLogger, format } from "@mangrovedao/commonlib.js";
+import {
+  createLogger,
+  CommonLogger,
+  format,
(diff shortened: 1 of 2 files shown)


packages/mangrove.js/src/util/logger.ts
```diff
@@ -7,13 +7,7 @@ const stringifyData = (data) => {
...
 };
 
-// FIXME: Temporary copy until issue #220 is fixed
-export const logdataLimiter = (data: Record<string, any>): any => {
-  return inspect(data, { maxStringLength: 1000 });
-};
-
-// FIXME: Temporary dumb toggle until issue #220 is fixed
-let loggingEnabled = true;
+let loggingEnabled = false;
```
espendk, line 10: This flag isn't used by anything?

espendk (changes requested):

espendk (approved):
Just tried it locally in a node REPL and it seems to work 👍

PR #489, opened 2022-06-01
fix: repair loglevel attribution (part of #220)

adhusson:


espendk (commented):
I think the actual commit looks fine: b09d9d6109f7518ff6adffd07697ff78e01d92ea

But I’m a little concerned about the diff in this PR, which contains a lot more (due to the reuse of a previously squash merged branch). Personally, I’d move the commit to a new branch to avoid the risk of messing up master :sweat_smile:

PR #531, opened 2022-06-28
feat!: Add Mangrove address to relevant strat events

espendk:
All strat events that pertain to a particular Mangrove instance now include the Mangrove address.


jkrivine (commented):
Looks all good to me. An alternative we might consider (to check whether that makes proxima's life complicated) is not to add mgv to those events, but require each maker contract being deployed (to be added in `MangroveOffer`'s constructor) to emit a `DeployedOn(IMangrove mgv)` log. This way you can always tell which Mangrove contract a log emitter is acting on

espendk:
Hmm, couldn't it be the case that a deployed strat could be reconfigured to use a different Mangrove? Or multiple Mangroves?

Unless the gas consumption is significant, I think I prefer including it in each of the events such that they're independently meaningful.

jkrivine:
> Hmm, couldn't it be the case that a deployed strat could be reconfigured to use a different Mangrove? Or multiple Mangroves?

not using the strat librairies (mgv is an immutable). 
 
> Unless the gas consumption is significant, I think I prefer including it in each of the events such that they're independently meaningful.

I don't think the gas cost is a lot though/

espendk:
Ok. For the sake of (hopefully) getting this deployed and adopted by Proxima this week (before my holiday), I suggest we go ahead with this PR.

jkrivine:
agreed

PR #541, opened 2022-07-01
Feature/foundry

adhusson:
**Note** git global config to automatically recurse when cloning, pulling, etc: `git config --global submodule.recurse true`

This seems ready for merging. It includes
* Nicer testing tooling that before built around forge tests and [forge-std](https://github.com/foundry-rs/forge-std/)
* Move solidity tests to foundry -- lots of test cleanup while translating to foundry, they should be used as examples of how to write tests
* Move all js tests to foundry (including Mangrove's permit functionality which had not been tested for a while)
* Use a custom script instead of hardhat to export ABIs/bytecodes
* Hardhat still compiles and generally works (see `hardhat-remappings.txt` and the `hardhat-preprocessor` plugin)
* Included foundry Github actions (to be improved)
* reorganised `strategies` contracts -- related to this, there are `TODO-foundry-merge` markers to grep and fill in, paging @jkrivine.

Note that you will need to `git submodule update --init --recursive` after switching to the branch. And you will need to `curl -L https://foundry.paradigm.xyz | bash` then `foundryup` to install foundry.

My intended next steps after merging are 
1) present foundry to you
2) gradually remove all dependencies on hardhat
3) improve github actions

We can also use this pull request to discuss those steps!


adhusson:
current status: I'm moving all JS tests to solidity to avoid dealing with the memory issue problem in github workflow.

PR #565, opened 2022-07-27
Feature/commenting

jkrivine:


packages/mangrove-solidity/contracts/test/strategies/Guaave.t.sol
```diff
@@ -95,13 +95,18 @@ contract GuaaveTest is MangroveTest {
...
+    tokens[0] = weth;
+    tokens[1] = usdc;
+
+    vm.expectRevert("Router/NotApprovedByMakerContract");
+    mgo.checkList(tokens);
+
+    // no need to approve for overlying transfer because router is the reserve
+    mgo.activate(tokens);
+    try mgo.checkList(tokens) {} catch Error(string memory reason) {
+      fail(reason);
```
adhusson, line 108: why not let the test fail? ie just `mgo.checkList(tokens)`

PR #566, opened 2022-07-27
Unit tests/strats

jkrivine:


packages/mangrove-solidity/contracts/strategies/single_user/abstract/SingleUser.sol
```diff
@@ -51,7 +51,7 @@ abstract contract SingleUser is MangroveOffer {
     if (!has_router()) {
       return TransferLib.transferToken(IERC20(token), receiver, amount);
     } else {
-      router().withdrawToken(token, reserve(), receiver, amount);
+      return router().withdrawToken(token, reserve(), receiver, amount);
```
adhusson, line 54: since return statements are used, shouldn't we have the signature be `returns (bool)` (unnamed)?

packages/mangrove-solidity/contracts/test/strategies/MangroveOffer.t.sol
```diff
@@ -0,0 +1,214 @@
...
+      "Incorrect default gasreq for simple maker"
+    );
+  }
+
+  function test_CheckList() public {
+    IERC20[] memory tokens = new IERC20[](2);
+    tokens[0] = weth;
+    tokens[1] = usdc;
+    vm.expectRevert("MangroveOffer/AdminMustApproveMangrove");
+    makerContract.checkList(tokens);
```
adhusson, line 65: maybe could use some love from [dynamic](https://github.com/mangrovedao/mangrove/blob/master/packages/mangrove-solidity/contracts/test/lib/Utilities.sol#L180) adapted to `IERC20`, so one can just write `makerContract.checkList(dynamic([weth,usdc]))`

packages/mangrove-solidity/contracts/test/strategies/MangroveOffer.t.sol
```diff
@@ -0,0 +1,214 @@
...
+    vm.prank(maker);
+    makerContract.approveMangrove(weth);
+    makerContract.approveMangrove(usdc);
+    // after approval, checkList should no longer revert
+    try makerContract.checkList(tokens) {} catch Error(string memory reason) {
+      fail(reason);
```
adhusson, line 71: I think just calling checkList should work, and the error will be displayed in test results

packages/mangrove-solidity/contracts/test/strategies/MangroveOffer.t.sol
jkrivine: Not sure how to do this (overloading `dynamic` with `dynamic(IERC20[2])` sounds a bit overkill?

adhusson: Stuffing `Utilities.sol` full of ease-of-life functions that makes the test super clear seems OK to me! Actually it's just copy-paste so we could preprocess those for any needed type at a later point.

packages/mangrove-solidity/contracts/test/lib/MangroveTest.sol
```diff
@@ -396,4 +406,14 @@ contract MangroveTest is Test2, HasMgvEvents {
   function $(IERC20 t) internal pure returns (address payable) {
     return payable(address(t));
   }
+
+  function tkn_pair(IERC20 t, IERC20 s)
```
adhusson, line 410: This is harder to remember than just updating the generic `dynamic`, I think just extending the existing static -> dyn conversion so we can do `fn(dynamic([tkn1,tkn2]))` would be easier for users/readers of the testing framework.

packages/mangrove-solidity/contracts/test/lib/MangroveTest.sol
```diff
@@ -41,6 +41,12 @@ contract MangroveTest is Test2, HasMgvEvents {
     uint defaultFee;
   }
 
+  modifier prank(address a) {
```
adhusson, line 44: I would caution against hiding testing details in modifiers, helper functions etc -- just the `cash` function gave me a lot of headache 😅 (and I'm worried it actually makes tests harder to read). Things like "calling a cheatcode" that break the normal EVM semantics should be very explicitly visible in tests so reviewers of a failing test know what's going on.

packages/mangrove-solidity/contracts/test/lib/MangroveTest.sol
```diff
@@ -361,9 +370,10 @@ contract MangroveTest is Test2, HasMgvEvents {
 
   /* **** Token conversion */
   /* return underlying amount with correct number of decimals */
-  function cash(TestToken t, uint amount) public returns (uint) {
+  function cash(TestToken t, uint amount) public view returns (uint) {
```
adhusson, line 373: BTW this will need conflict resolution since I merged #568

packages/mangrove-solidity/contracts/test/strategies/integration/Mango.t.sol
```diff
@@ -236,14 +231,16 @@ contract MangoTest is MangroveTest {
     // - verify takerGave is pending
     // - put back the density and run another market order
     mgv.setDensity($(weth), $(usdc), cash(weth, 1));
-    vm.prank(taker);
```
adhusson, line 239: I don't understand these changes (unless they were made before the `cash`/`savePrank` update?)

adhusson (commented):
Super cool way to have tests of multiple implementations of the same interface in one go!!

Note: in addition to comments above, I think there's a missing `try...` to eliminate in [Guaave.t.sol](https://github.com/mangrovedao/mangrove/blob/97c10bfd37ae3d566b2065133a286b6897cab432/packages/mangrove-solidity/contracts/test/strategies/integration/Guaave.t.sol#L107)

packages/mangrove-solidity/contracts/test/lib/MangroveTest.sol
jkrivine: OK but modifier is quite explicit no?

packages/mangrove-solidity/contracts/test/strategies/integration/Mango.t.sol
jkrivine: yes

packages/mangrove-solidity/contracts/test/lib/MangroveTest.sol
adhusson: Not enough I think -- you would read it and wonder why `startPrank`/`stopPrank` wasn't used, read the definition, then realize it's because in this specific case the entire function code could be pranked instead of just a subpart of it...

PR #570, opened 2022-08-02
fix(build)!: Change mangrove-solidity local tests to use .env.local.test

lnist:
Also upped the available memory for JavaScript heap when building mangrove.js.

Consequential changes from https://github.com/mangrovedao/mangrove/pull/541 related to the epic https://github.com/mangrovedao/mangrove-tasks/issues/129


lnist:
@adhusson : We're merging this to fix running tests locally, but can you verify that it aligns with what you had in mind when sourcing .env from bash?

adhusson:
it looks good, I did see something a little more general at this url https://github.com/entropitor/dotenv-cli/issues/37 but not a huge difference

Issue #113, opened 2022-01-10
add `snipe` function to `market.ts`


jkrivine:
I suggest either to add a specific `snipe` function or add optional argument `offerId` to `buy` function. If `offerId` is defined then `buy` would trigger a `snipe` on mangrove, otherwise proceed with the `marketOrder`

jkrivine:
Mangrove accepts two kinds of taker order: a market order or a snipe. In the latter case, one targets specific offerIds instead of relying on the matching engine to select offers.
In the backend this is implemented by calling:
```solidity
function snipes(
    address outbound_tkn,
    address inbound_tkn,
    uint[4][] calldata targets,
    bool fillWants
  )
    external
    returns (
      uint successCount,
      uint snipesGot,
      uint snipesGave,
      uint totalPenalty,
      uint feePaid
    )
    ```
    One should provide a way to snipe offers in the API either by implementing a `snipe` function or by adding a parameter to `buy` and `sell` existing functions.
   Use case for sniping is:
   - sniping a failing offer to collect the bounty
   - homemade market order taking gas cost into account

PR #587, opened 2022-08-10
Add offerId to buy/sell

lnist:
Fixes part of #113.

This adds `offerId` to buy/sell to enable choosing a specific order while also using the volume/price/etc. calculation in buy/sell.

A second PR will add general snipes/cleaner in the API.

The gas estimation for snipes has currently been implemented to either choose _one_ value from override or estimate based on each target's wants values. Do you agree?

partialFill is simply set to false - it was not clear to me what it should be set to when making the event summary for multiple targets. Let me know if I should change it to be exactly like for market order.


packages/hardhat-utils/config/hardhat-mangrove-config.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
         gasMultiplier: 1,
         blockGasLimit: 7000000000,
         allowUnlimitedContractSize: true,
+        loggingEnabled: false,
```
dontrolle, line 12: Hmmm, I guess it's ok to explicitly disable logging as the default. Makes it easier to change it to true, when needed. 👍

dontrolle (approved):
Looks good to me 👌

PR #590, opened 2022-08-10
Feature/peter/252

peterMangrove:
I've create 3 classes for trade events, trades and unitCalculations. I've extracted some logic from market, semibook and mangrove, to these 3 classes. This is to keep the size of market, semibook and mangrove down, make them easier to understand and to make some of the logic more testable.

The API is still kept as is in alle classes, the methods now simply call the logic classes. For example market still has the method buy, but it simple calls the trade class, which holds the logic for a buy action. By doing this, the size of market is kept down and for new people, it will hopefully make more sense that the buy logic is in the Trade class.

One problem with the structure of the code today, in for example market, is that there is a lot of private methods and because of this, it is not possbile to make unit tests. A way to fix this, would be to move private methods in to a suiting logic class and make them public. This way the API stays the same and the methods become more testable.

I've create unit tests for all 3 classes. I've created a new folder for unit tests, so that they are not mixed with our integration test. In doing so, i have added af package called mockito-ts. This is a mocking tool, that makes easier to unit test classes, the documentation can be found here https://github.com/NagRock/ts-mockito.

I've changed til package.json, so that it is now possible to run the unit test without spinning up hardhat. This means that you can run unit tests by writing "yarn test:unit" and integration test with "yarn test:integration". "yarn test" will run both commands. This makes the unit test run alot faster

Last thing is, that i've added some more integration test for both market and semibook, for the methods that had no tests.
In market the methods "onceWithTxPromise" and "unsubscribe", still isn't testet fully.


packages/mangrove.js/src/mangrove.ts
```diff
@@ -267,14 +270,9 @@ class Mangrove {
    *  mgv.toUnits(10,6) // 10e6 as ethers.BigNumber
    *  ```
    */
+  //TODO: maybe move to MgvToken
```
espendk, line 273: remove comment

packages/mangrove.js/src/mangrove.ts
```diff
@@ -399,6 +388,7 @@ class Mangrove {
    * Read decimals for `tokenName` on given network.
    * To read decimals directly onchain, use `fetchDecimals`.
    */
+  //TODO: move to MgvToken
```
espendk, line 391: this too

packages/mangrove.js/src/util/tradeEventManagement.ts
```diff
@@ -0,0 +1,110 @@
...
+    ) => boolean
+  ): Market.Summary {
+    return {
+      got: got.fromUnits(event.args.takerGot),
+      gave: gave.fromUnits(event.args.takerGave),
+      partialFill: partialFillFunc(event.args.takerGot, event.args.takerGave),
+      penalty: this.mangroveUtils.fromUnits(event.args.penalty, 18),
+    };
+  }
+  createSummaryForOrderComplete(
```
espendk, line 42: could be nice to have consisten naming, ie *FromEvent

packages/mangrove.js/src/util/tradeManagement.ts
```diff
@@ -0,0 +1,129 @@
+import Big from "big.js";
+import Market from "../market";
+import MgvToken from "../mgvtoken";
+import { Bigish } from "../types";
+import UnitCalculations from "./unitCalculations";
+
+class TradeManagement {
```
espendk, line 7: maybe rename to TradeParameters or (if buy, sell, etc is also moved here) TradeExecution

packages/mangrove.js/src/util/tradeManagement.ts
```diff
@@ -0,0 +1,129 @@
...
+  mangroveUtils = new UnitCalculations();
+
+  //TODO: Semibook related
```
espendk, line 10: delete

espendk (changes requested):

espendk (approved):

PR #592, opened 2022-08-11
fix(test): Don't override NODE_OPTIONS's vscode bootloader

lnist:
VSCode appends a --require /very/long/path/extensions/ms-vscode.js-debug/src/bootloader.bundle.js to NODE_OPTIONS which I accidentally overrode when increasing JS heap memory. Now we no longer override but use cmd line options instead.


espendk (approved):
good find! lgtm

PR #593, opened 2022-08-11
fix(build): Better `find references` in vscode

lnist:
Improves tsconfig files and adds a global file referencing all projects to improve code navigation.

For instance looking at references to `Market` class now displays all the bots. However, since, e.g., `mgvIntegrationTestUtil.ts` is not in a typescript project as such and simply referenced by tests, the find references does not work well. If, however one first looks at references to something in production code, then it seems vscode causes typescript to load all projects and then it also works for the utils. We should consider moving utils into its own package which tests can then reference.

Next improvement step would be to consider to use yarn PnP with its typescript plugin.

Note:
1) mgv cli was already broken due to chalk being upgraded. Can probably be bumped when we upgrade to TypeScript 4.7.
2) tsconfig change broke mgv cli since it is outside ts root (./src), fixed by using node imports for package.json.
2) Build has slightly different output in mangrove.js dist. I assume due to changed ts settings.
3) Also d.ts files are generated for more projects due to the ts defaults for composite: true.
4) Also applies fix like c36e85b7eabc04f4e9ec7172ffe0d774f7d6e704 but for bot integration tests.


espendk (approved):
awesome

PR #598, opened 2022-08-14
Feature/unified deploy

adhusson:
Would like to know your opinion on this change to deploys management.

Current strategy under foundry is to parse "foundry broadcast logs". They are files that summarize an onchain deployment.

Pros:
* Managed by foundry
* Separate unsent txs from mined txs

Cons:
* Too generic to e.g. easily parse a deploy that creates >1 instance of the same contract
* Not trivial to parse

Proposed strategy here is to write a deployment file as part of the deployment.
Pros:
* Can write arbitrary data to the deployment file, give special names to deployed contracts
* End result is immediately readable by e.g. mangrove.js

Cons:
* Ad-hoc, does not reuse existing foundry infrastructure.
* File is generated when the script is executed but not after the transactions have been mined, so could create a discrepancy between what's in the deployment files and what's actually onchain.

P.S. Ignore the renamed files, changes start here: https://github.com/mangrovedao/mangrove/compare/feature/anviltests...feature/unifiedDeploy#diff-0e0a0b427ef3af9b716c4c35566e7a95310881d5018b28398a5d30c539528d0c

### DIFF
--- a/packages/mangrove-solidity/script/MangroveJs.s.sol
+++ b/packages/mangrove-solidity/script/MangroveJs.s.sol
@@ -53,7 +53,7 @@ contract MangroveJsDeploy is Deployer {
       symbol: "TokenA",
       _decimals: 18
     });
-    ens.set("TokenA", address(tokenA), true);
+    register("TokenA", address(tokenA), true);
 
     vm.broadcast();
     tokenB = new TestToken({
@@ -62,7 +62,7 @@ contract MangroveJsDeploy is Deployer {
       symbol: "TokenB",
       _decimals: 18
     });
-    ens.set("TokenB", address(tokenB), true);
+    register("TokenB", address(tokenB), true);
 
     vm.broadcast();
     dai = new TestToken({
@@ -71,7 +71,7 @@ contract MangroveJsDeploy is Deployer {
       symbol: "DAI",
       _decimals: 18
     });
-    ens.set("DAI", address(dai), true);
+    register("DAI", address(dai), true);
 
     vm.broadcast();
     usdc = new TestToken({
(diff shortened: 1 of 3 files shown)


jkrivine:
Looks great. I'll review asap

espendk:
I'll start reviewing now :-)

This stated con for the proposal is quite significant though:
> - Ad-hoc, does not reuse existing foundry infrastructure.

We'll be on our own wrt bugs, maintenance and new features, and any improvements to the Foundry approach&tooling won't benefit us.

packages/mangrove.js/src/util/getMangroveAddresses.ts
```diff
@@ -1,46 +1,28 @@
...
-  contract SubA is A {}
+Update deployment addresses using mangrove-solidity deployment data.
 
+Writes a new file, if output file is not already present. 
 
 Args:
+  [--debug]
 
   [--debug] Add --debug flag to get debug output.
   [--noop] Do not actually update addresses.json.
```
espendk, line 10: This option no longer seems to be supported.

espendk (commented):
I don't have a strong opinion here, as the differences between the two approaches don't seem that big - the devX seems pretty much the same? I'm probably missing something as I'm not (yet) a Solidity dev 😅

As mentioned previously, I'd be cautious not to follow Foundry's best practices/approach _if_ there are such.

adhusson:
> I don't have a strong opinion here, as the differences between the two approaches don't seem that big - the devX seems pretty much the same? I'm probably missing something as I'm not (yet) a Solidity dev 😅
> 
> As mentioned previously, I'd be cautious not to follow Foundry's best practices/approach _if_ there are such.

Yes the devX is the same -- it just makes mangrove-solidity produce cleaner, easier to consume files, requires less logic in mangrove.js, but has the big con as you highlighted of not using foundry broadcast logs. I'll pause this for now and just import some tangentional improvements of the branch into `feature/anviltests`

adhusson:
Note: I've incorporated most changes, only thing remaining on this branch is the "write a custom json file on deployment execution, even if not broadcasting".

adhusson:
Closing as it was superseded by #719

PR #601, opened 2022-08-15
feat(bot-updategas): update to use v2 of Mumbai gasstations

peterMangrove:
Updated the config to use v2 of Mumbai gasstations.
Changed the logic of GasUpdater to be able to use both an oracleKey and oracleSubKey. This is necessary, because the json schema for the gasstation has changed. And the standard gasprice, now has 2 options, "maxFee" and "maxPriorityFee". The subKey is now "maxFee"

All the private methods from GasUpdater, has been moved to a new class "GasHelper". This made i possible to unit test all these methods. There is now unit test for all methods in GasHelper.

The git workflow for the gasUpdater bot, has been updated to run integration test and unit test separately


dontrolle (approved):
👍

PR #605, opened 2022-08-16
feat(snipe): Add ability to snipe a list of targets

lnist:
Exposes snipe on mangrove.js - partially fixes https://github.com/mangrovedao/mangrove/issues/113

See also https://github.com/mangrovedao/mangrove-docs/pull/21


packages/mangrove.js/src/market.ts
```diff
@@ -80,6 +81,17 @@ namespace Market {
     provision: Bigish;
   };
 
+  export type SnipeParams = {
+    targets: {
+      offerId: ethers.BigNumberish;
```
espendk, line 86: It looks like we use `number` for offerIds in the code above, so should probably be the same here.

packages/mangrove.js/src/market.ts
```diff
@@ -80,6 +81,17 @@ namespace Market {
...
+      takerWants: Bigish;
+      takerGives: Bigish;
+      gasLimit?: ethers.BigNumberish;
```
espendk, line 89: We use `number` for `gasForMarketOrder` above, so should probably be the same here.

packages/mangrove.js/src/market.ts
```diff
@@ -499,6 +511,23 @@ class Market {
...
 
+  /**
+   * Snipe specific offers.
+   * Params are:
+   * `targets`: an array of
+   *    `offerId`: the offer to snipe
+   *    `takerWants`: the amount of base token (buy order) or quote token (sell order) the taker wants
+   *    `takerGives`: the amount of quote token (buy order) or base token (sell order) the take gives
+   *    `gasLimit?`: the maximum gas requirement the taker will tolerate for that offer
+   * `orderType`: whether to use the base/quote (buy) or quote/base (sell) market
```
espendk, line 522: This phrasing is weird to me - there are not two markets, only one.
Maybe we could write: "whether to snipe 'asks' or 'bids'". Maybe the parameter options should actually be "asks"/"bids"

packages/mangrove.js/src/market.ts
```diff
@@ -499,6 +511,23 @@ class Market {
...
+   * `fillWants?`: fillWants specifies whether you are acting as a buyer of outbound tokens, in which case you will buy at most takerWants, or a seller of inbound tokens, in which case you will buy as many tokens as possible as long as you don't spend more than takerGives.
```
espendk, line 523: Not sure the phrasing "acting as a .." makes sense to me, so cut that out:

* `fillWants?`: specifies whether you will buy at most `takerWants` (true), or you will buy as many tokens as possible as long as you don't spend more than `takerGives` (false).

packages/mangrove.js/src/util/trade.ts
```diff
@@ -252,6 +252,53 @@ class Trade {
...
+  /**
+   * Snipe specific offers.
+   * Params are:
+   * `targets`: an array of
+   *    `offerId`: the offer to snipe
+   *    `takerWants`: the amount of base token (buy order) or quote token (sell order) the taker wants
+   *    `takerGives`: the amount of quote token (buy order) or base token (sell order) the take gives
+   *    `gasLimit?`: the maximum gas requirement the taker will tolerate for that offer
+   * `orderType`: whether to use the base/quote (buy) or quote/base (sell) market
+   * `fillWants?`: fillWants specifies whether you are acting as a buyer of outbound tokens, in which case you will buy at most takerWants, or a seller of inbound tokens, in which case you will buy as many tokens as possible as long as you don't spend more than takerGives.
```
espendk, line 264: As above.

packages/mangrove.js/src/util/trade.ts
```diff
@@ -252,6 +252,53 @@ class Trade {
...
+   */
+  snipe(
+    params: Market.SnipeParams,
+    overrides: ethers.Overrides = {},
+    market: Market
+  ): Promise<Market.OrderResult> {
+    const fillWants = "fillWants" in params ? params.fillWants : true;
```
espendk, line 271: Maybe the nullish coalescing operator would be more concise:
```
const fillWants = params.fillWants ?? true
```

packages/mangrove.js/test/integration/market.integration.test.ts
```diff
@@ -890,6 +890,184 @@ describe("Market integration tests suite", () => {
     expect(result.successes[0].offerId).to.be.equal(notBest);
   });
 
+  it("snipe asks book for two successful orders succeeds", async function () {
```
espendk, line 893: I don't see `feePaid` - was it impossible to test?

packages/mangrove.js/test/unit/semibook.unit.test.ts
```diff
@@ -1,7 +1,5 @@
 import assert = require("assert");
-import { Semibook } from "../../src";
-import Market from "../../src/market";
-
+import { Semibook, Market } from "../../";
```
espendk, line 2: This uses the compiled versions of Semibook and Market, right? For unit tests, maybe we should use the source code versions instead? (this would also match the following changes which do exactly that for other test files).

espendk (changes requested):
nice. A few minor comments, but otherwise lgtm

packages/mangrove.js/src/market.ts
lnist: Agree, I've gone with both your suggestions. The asks/bids change had some additional consequences, see commit.

packages/mangrove.js/src/market.ts
lnist: more concise, I like it.

packages/mangrove.js/src/util/trade.ts
lnist: Good suggestion - It also explicitly converts null to true; and since true is meant to be the default then that is better.

packages/mangrove.js/test/integration/market.integration.test.ts
lnist: I thought it would have far-reaching consequences for tests if fees were raised. But only a few changes were needed, so I've done it :)

packages/mangrove.js/test/unit/semibook.unit.test.ts
lnist: We flipped back and forth locally trying to fix build errors. I still think vscode gets confused sometimes (in integration tests) and I am not entirely sure what happens. But here I agree it should point to src.

espendk (approved):
lgtm

The new non-zero fee in integration tests should probably be announced to the dev team so people are aware of it.

PR #607, opened 2022-08-16
feat(bot-updategas): added option to set constraints on the max change to the gas price

peterMangrove:
This adds an option to the update gas bot, where you can set an max value the gas price can change.

It can be done with either a constant and/or a percentage of the current price.

Added unit test for the logic and integration tests.


packages/bot-updategas/README.md
```diff
@@ -51,9 +51,14 @@ Here's an example configuration file with instances of all possible configuratio
   "logLevel": "info",
   "acceptableGasGapToOracle": 0.5,
   "constantOracleGasPrice": 3,
-  "oracleURL": "https://gasstation-mainnet.matic.network/",
+  "oracleURL": "https://gasstation-mainnet.matic.network/v2",
```
dontrolle, line 54: Good catch updating the docs for this and the previous. ;)

packages/bot-updategas/src/GasUpdater.ts
```diff
@@ -126,10 +146,31 @@ export class GasUpdater {
...
+          data: { allowedNewGasPrice },
         });
+        const [isAllowed] = this.gasHelper.shouldUpdateMangroveGasPrice(
+          currentMangroveGasPrice,
+          allowedNewGasPrice,
+          this.#acceptableGasGapToOracle
+        );
+        if (!isAllowed) {
+          logger.error(
+            "The max update constraint is lowering/increasing the gas price, so that it is within the the acceptableGasGap"
```
dontrolle, line 167: Minor: Double "the" in the log message.

dontrolle (commented):
Looks very good. Very good with a number of tests in the new scaffolding you introduced. Apart from the minor comments, let's get this merged.

PR #610, opened 2022-08-16
feat: added workflow step, to added Coveralls

peterMangrove:
POC for code coverage

It is now possible to se the overall code coverage of the repo. 
This is done by creating  lcov report for every package
The github workflow then uploads the reports to coveralls.io, where it is possible to see more stats of what code is tested.

The readme.md for the repo has been updated to show a badge of the code coverage percentage.


packages/mangrove.js/package.json
```diff
@@ -67,6 +67,7 @@
     "async-mutex": "^0.3.2",
     "big.js": "^6.1.1",
     "config": "^3.3.7",
+    "coveralls": "^3.1.1",
```
espendk, line 70: I think this should be a devDependency instead as it will only be used by CI.

How come this is only in the mangrove.js package?

README.md
```diff
@@ -1,5 +1,7 @@
 [![CI](https://github.com/mangrovedao/mangrove/actions/workflows/node.js.yml/badge.svg)](https://github.com/mangrovedao/mangrove/actions/workflows/node.js.yml)
 
+[![Coverage Status](https://coveralls.io/repos/github/mangrovedao/mangrove/badge.svg)](https://coveralls.io/github/mangrovedao/mangrove)
```
espendk, line 3: Would be nice to have it next to the icon above

packages/bot-maker-noise/test/unit/offerMaker.unit.test.ts
```diff
@@ -0,0 +1,5 @@
+import { describe, it } from "mocha";
+
+describe(" Unit test suite for OfferMaker", () => {
```
dontrolle, line 3: Is this file (and similar files in other dirs) commited now to make it easier to fill out tests later? If so - I'm fine with it. :)

dontrolle (commented):
A minor question (see below), but apart from that lgtm. :) 

Great that we can begin to track coverage!

PR #621, opened 2022-08-17
feat(mangrove.js): generalize volume estimations (#620)

### LINKED ISSUE #620: mangrove.js: Add price limit in volume estimation
Per @johanleroch, mangrove.js could use a volume estimation that includes an average price limit.

adhusson:
Closes #620.

Summary: mangrove.js now faithfully simulates mangrove's solidity execution engine. The `estimateVolume` function has been changed and includes a new `boundary` parameter.

If `params.to == buy`, then the mapping to solidity code is:
* `params.given` --> `takerWants`
* `params.boundary` --> `takerGives`

Otherwise if `params.to == sell`, then the mapping is:
* `params.given` --> `takerGives`
* `params.boundary` --> `takerWants`

You can omit the `estimateVolume`'s `boundary` param and use it like before. It will automatically set a very high max avg. price when buying, and a 0 min avg. price when selling.

### DIFF
--- a/packages/mangrove.js/src/semibook.ts
+++ b/packages/mangrove.js/src/semibook.ts
@@ -26,14 +26,18 @@ namespace Semibook {
    * Specification of how much volume to (potentially) trade on the semibook.
    *
    * `{given:100, to:"buy"}` means buying 100 base tokens.
+   * `{given:100, to:"buy", boundary: 10})` means buying 100 quote tokens for a max. avg. price of 1/10 (boundary/given).
    *
    * `{given:10, to:"sell"})` means selling 10 quote tokens.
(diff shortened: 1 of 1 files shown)


johanleroch:
should it also change something for the functions estimateVolumeToSpend and estimateVolumeToReceive?

adhusson:
> should it also change something for the functions estimateVolumeToSpend and estimateVolumeToReceive?

good catch, I just checked and thanks to good type layout those functions got implicitly updated (they now have a new `boundary` parameters field available).

johanleroch:
> > should it also change something for the functions estimateVolumeToSpend and estimateVolumeToReceive?
> 
> good catch, I just checked and thanks to good type layout those functions got implicitly updated (they now have a new `boundary` parameters field available).

Nice, good job so :)

jkrivine (approved):
Looks all good to me

PR #629, opened 2022-08-19
Feature/marketclean

lnist:
Preparation for cleaner bots, follow-up PR will change our example bot-cleaning to use the implementation.
See https://github.com/mangrovedao/mangrove/issues/113


packages/mangrove.js/test/integration/semibook.integration.test.ts
```diff
@@ -243,11 +243,13 @@ describe("Semibook integration tests suite", () => {
   describe("getBestInCache", () => {
     it("returns undefined, because market made before offer", async function () {
       const market = await mgv.market({ base: "TokenA", quote: "TokenB" });
+      market.getSemibook("asks").disconnect();
```
espendk, line 246: How come the semibook should be disconnected? I think this breaks the idea of the test as we're no longer observing how the semibook responds to events (it doesn't receive any?)

espendk (commented):
nice improvement :-) One question though re. disonnecting a semibook, see below.

lnist:
@espendk : Do you have an opinion on the parseLogs question above? and does the other answer resolve the conversation?

espendk:
> @espendk : Do you have an opinion on the parseLogs question above? and does the other answer resolve the conversation?

I don't see any questions on this page at all, except the one I asked re. disconnecting the semibook?

packages/mangrove.js/test/integration/semibook.integration.test.ts
espendk: There's a synchronization barrier before checking the cache, so there shouldn't be a race condition.
So I see three possibilities:
1. the synchronization barrier isn't working as it should
2. the test is wrong, ie `getBestInCache` should not return `undefined` after an offer has been added. Note though that the cache is partial and events don't carry sufficient information to always keep the cache full, so it's not obvious to me that the new offer should be in the cache
3. there's a bug

I don't know what the answer is, that requires some investigation. But I'm quite sure that disconnecting the semibook makes the test pointless.

packages/mangrove.js/src/util/trade.ts
```diff
@@ -560,7 +565,19 @@ class Trade {
...
+    const snipeFunction = requireOffersToFail
+      ? market.mgv.cleanerContract.collect
+      : market.mgv.contract.snipes;
+    const parseLogs = requireOffersToFail
+      ? (events: ethers.Event[], logs: ethers.providers.Log[]) =>
+          logs
+            .filter((x) => x.address === market.mgv._address)
+            .map((l) => market.mgv.contract.interface.parseLog(l))
+      : (events: ethers.Event[], logs: ethers.providers.Log[]) =>
+          events.filter((x) => x.address === market.mgv._address);
```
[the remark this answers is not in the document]
espendk: I don't think there's any other way to do it. I think the reason is that the contract object you're using to send the tx only knows the ABI of the cleaner contract and thus it can only parse events from that, so it only populates `events` with cleaner contract events. 

But @adhusson would know better :-)

espendk (changes requested):

packages/mangrove.js/test/integration/semibook.integration.test.ts
lnist: I've changed it to use the new sync mechanism instead and asserting that the book _does_ get updated in the test.

PR #630, opened 2022-08-19
CI: added workflow check for onchain abis

### LINKED ISSUE #259: Add CI check on `master` that mangrove.js ABI's match ABI's of deployed contracts
**Is your feature request related to a problem? Please describe.**
The root cause of issue #256 was that the ABI for Mangrove had changed on `master` (including in mangrove.js), but the new version of the contract hadn't been deployed. This meant that mangrove.js couldn't understand the events of the Mangrove contract instance it was connecting to.

**Describe the solution you'd like**
There should be a CI check on `master` which verifies that the ABI's in mangrove.js match those on the relevant block explorers or (for Mangrove contracts) those in the `deployments` folder of `mangrove-solidity`:

- The ABI's contained in mangrove.js will be part of the checked out code in CI, e.g. https://github.com/mangrovedao/mangrove/blob/master/packages/mangrove.js/src/abis/Mangrove.json
- The ABI's for Mangrove contract deployments will also be part of the checked out code in CI, e.g. https://github.com/mangrovedao/mangrove/blob/master/packages/mangrove-solidity/deployments/mumbai/Mangrove.json
- The ABI for the verified contract can be fetched from PolygonScan using their API: [docs.polygonscan.com/api-endpoints/contracts#get-contract-abi-for-verified-contract-source-codes](https://docs.polygonscan.com/api-endpoints/contracts#get-contract-abi-for-verified-contract-source-codes). Similar API's should exist for other block explorers.

Comparing the ABI's can probably be done using something like `fast-equals` or `lodash`.

**Describe alternatives you've considered**
A warning when ABI files change stating that one must remember to redeploy Mangrove.

**Additional context**
None

peterMangrove:
Added code, that checks if the Mangrove abi onchain matches the abi locally in the mangrove.js

This is done by getting the abi from polygon testnet and comparing it with the abi locally in mangrove.js

The abi onchain has slightly different types, because of this 4 types are replaced with the correct type. Look in "compareOnChainAbi.ts". This file creates a file called "AbiAreEqual.json", if the abi onchain is equal to the one locally.

The CI workflow runs this file and checks if the file is created. If the file is not created it shows a warning, that the abi onchain isn't matching the abi locally.

fixes #259


adhusson:
Archiving for now, see #259.

Issue #148, opened 2022-01-19
`LiquidityProvider` should provide a `consoleAsks/Bids` function


jkrivine:
Right now one can do `lp.market.consoleAsks()` which pretty prints the cached `asks` of the market. Calling `lp.consoleAsks()` should do the same but restricted to the asks posted by the liquidity provider.

dontrolle:
This issue seems like a **good first issue** for newcomers, but could maybe do with a bit more spec about what you'd like, @jkrivine to help. :)

jkrivine:
I can't add much more than the above succinct description. The `market` class provides a `consoleAsks` and `consoleBids` function to pretty print the list of offers in the order book cache. 
I suggest that the `liquidityProvider` class provides also a similar function but this time restricted to offers owned by `liquidityProvider.logic`

PR #637, opened 2022-08-22
feat(mangrove.js): added functionality for console-Ask and -Bids for an liquidityProvider

peterMangrove:
Moved logic for pretty print out to its own class.

Market's consoleAsks, consoleBids and prettyPrint now uses this new class to print offers

LiquidityProvider now also has consoleAsks and consoleBids, but only for the makers asks or bids.

Added unit test for pretty print.

fixes: #148


peterMangrove:
Hey @jkrivine Can you take a look at this PR, when you have time? :)

jkrivine (approved):
Looks good to me!

PR #650, opened 2022-08-24
doc(mangrove-solidity): minor fix to taker vs msg.sender

lnist:


adhusson (approved):
Thanks for the fix :)

PR #657, opened 2022-08-25
fix(bundling): Enable webapp to import mangrove.js via link

lnist:
See the individual commits and their headlines for details on the necessary fixes.


espendk (approved):
brilliant - that's a lot of stuff that needed to be fixed in the mangrove.js browser support 😅
And thanks for breaking the PR down into readable commits 🙏

packages/mangrove.js/package.json
```diff
@@ -11,7 +11,7 @@
     "precommit": "lint-staged",
     "prepack": "yarn run build",
     "lint": "npx eslint ./src/*.ts",
-    "build-this-package": "npm-run-all -pn --aggregate-output get-mangrove-abis get-mangrove-addresses && npm-run-all -pn --aggregate-output typechain lint && npm-run-all -pn --aggregate-output tscbuild esbuild && yarn make-cli-executable",
+    "build-this-package": "npm-run-all -pn --aggregate-output get-mangrove-abis get-mangrove-addresses && npm-run-all -pn --aggregate-output typechain lint && yarn tscbuild && yarn esbuild && yarn make-cli-executable",
```
adhusson, line 14: Can you expand on why you are running esbuild on the js export (and not the ts source)?

adhusson (commented):
Excellent work on improving the state of things browser-wise!

I'm wondering about 2 things:
* Running esbuild on js output. From looking around online it seems esbuild works hard to be ready to run on ts codebases, and parallelization is always nice.
* The file reorg seems to come down to having 2 files that must be shimmed (commonlib & readJsonWallet). But I'm not sure I understand what's going on. Anyway now the result seems to be a file structure that's driven by our build needs rather than just natural code organization (e.g. not having access to `node` in the mangrove.js exports), which will also inform all future developments. What do you think about either just shimming all node stuff that's necessary using [brode](https://github.com/geut/brode/tree/main/packages/esbuild-plugin-brode) ?

packages/mangrove.js/package.json
lnist: 1. "esbuild on js export": Building with esbuild on the ts source with iife modules created a mangrove.js file which our UI framework (vite) could not seem to use for building. It could not see any of our exports, so it could not find anything. "Mangrove", "Market" etc were all undefined. To debug it I ended up using the following to see what exports were in the files. With the old rollup-build we had a lot, but with the esbuild+ts+commonjs almost nothing is dumped (except something related to numbers):
   ```js
   const exportedVar = [];
   import('./node_modules/@mangrovedao/mangrove.js/dist/browser/mangrove.min.js')
    .then(module => {
        const {default: defaultExport, ...rest} = module;
        exportedVar.push(defaultExport.name, ...Object.keys(rest));
        console.log(exportedVar);
    });
    ```
    I admit that I don't know exactly why that is. Neither switching to js files or using commonjs alone fixed it. I had to do both. I am willing to give it another try now that it actually works overall :) . Also, the parallel build is nice, but the esbuild is so blazing fast that it doesn't really matter :)
    
2. I share your concern around code reorg. I did attempt to name the files around something functional, and I may have failed :D - IMHO a library to be used in a browser should not depend on node-functionality and hopefully we can find a separation of functionality for the browser-library and node-library, because if the browser does not support such functionality then why have it in the library? However, that may be too idealistic and something like `brode` would be a good bet. For now I suggest we keep it as is, and if we see node dependencies creep in again and again, then we can reconsider. My hope is that logging and file reading were the main issues.

adhusson: Thank you for finding a fix, hadn't realized I broke Vite compatibility which is super bad. Ok we can look at it more relaxed now :)

Agree on leaving as-is for now on the code org, I will take a quick look at esbuild if I find the time.

PR #659, opened 2022-08-25
[broken] WIP: namepace nothing so flattening can be done

adhusson:
Leaving a note here: this branch has the modifications necessary for Mangrove to be flattenable by regular tools (hardhat, foundry). It should be flattenable but current tools do not handle qualified imports at all. See this issue for foundry: foundry-rs/foundry/issues/2545.


adhusson:
Closing, too out of date now

PR #662, opened 2022-08-26
feat(snipe): Allow retrieval of raw parameters to send to contracts

lnist:
Add the ability to get raw parameter for sniping to use for callStatic and estimateGas methods.

We earlier decided to use number for max gas, that limits the max amount of gas we can pass in.

bot-cleaning is updated to use new api.

This completes #113  .

### DIFF
--- a/packages/bot-cleaning/src/MarketCleaner.ts
+++ b/packages/bot-cleaning/src/MarketCleaner.ts
@@ -1,7 +1,7 @@
 import { logger } from "./util/logger";
 import { Market, Semibook } from "@mangrovedao/mangrove.js";
 import { Provider } from "@ethersproject/providers";
-import { BigNumber, BigNumberish } from "ethers";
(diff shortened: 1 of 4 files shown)


packages/bot-cleaning/src/MarketCleaner.ts
```diff
@@ -13,7 +13,7 @@ type OfferCleaningEstimates = {
 
 // FIXME move to Mangrove.js
 // const maxWantsOrGives = BigNumber.from(2).pow(96).sub(1);
-const maxGasReq = BigNumber.from(2).pow(256).sub(1);
+const maxGasReq = Number.MAX_SAFE_INTEGER - 1;
```
dontrolle, line 16: Let's try a comment which requires a change.

dontrolle (changes requested):

dontrolle (approved):

PR #667, opened 2022-08-29
feat(bundling): Make esbuild use TypeScript when bundling

lnist:
Follow-up to https://github.com/mangrovedao/mangrove/pull/657

* Switch to iife bundling for esbuild - esbuild and rollup behave differently for some exports, and as commented in https://github.com/evanw/esbuild/issues/1182 we can work around this issue with a custom footer to avoid having to change the web-app's way of importing our bundle.
* Enable tsconfig flags which ensures esbuild works and make consequential changes.
* Switch to TypeScript for esbuild
* Run tscbuild and esbuild in parallel

### DIFF
--- a/packages/commonlib.js/src/index.ts
+++ b/packages/commonlib.js/src/index.ts
@@ -5,7 +5,7 @@ import { sleep } from "./promiseUtil";
 
 export {
   ErrorWithData,
-  CommonLogger,
+  type CommonLogger,
   createLogger,
   createConsoleLogger,
   logdataLimiter,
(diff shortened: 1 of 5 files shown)


adhusson (approved):
Looks great! Let's go. For my culture, the alternative was to make the webapp do something like `Mangrove = require("mangrove").default;`?

lnist:
> Looks great! Let's go. For my culture, the alternative was to make the webapp do something like `Mangrove = require("mangrove").default;`?
Yes, I believe so, but I haven't actually tried it :)

PR #671, opened 2022-08-29
NOO, DONT PUSH ME

espendk:
Test PR for requesting review from a team.

Currently, there's only one team containing all devs, but we can easily create more.
I think you can see your teams here: https://github.com/orgs/mangrovedao/teams

Please don't merge 🙏


lnist (changes requested):
Too few O's in NOO.

espendk:
Apparently, when one person does the review, the team is no longer listed as requested for review. Which makes sense, but means that people looking at this after @lnist's review didn't see it action, so I re-requested a review from the team.
But please don't actually do a review :-)

dontrolle:
@espendk Is this one still needed for some testing, or can we close it?

Issue #619, opened 2022-08-17
`LiquidityProvider`'s `newBid/Ask` event listener does not work


jkrivine:
**Describe the bug**
Using the API to post an offer works but the event filter is never satisified.

**To Reproduce**
1. `lp = await mgv.liquidityProvider(market)`
2. `await lp.newAsk({volume: 1, price: 2500, fund:0.1})`

succeeds but outputs "Uncaught { revert: false, exception: 'tx mined but filter never returned true' }" although the ask was indeed posted to Mangrove.

peterMangrove:
@jkrivine This seems to be related to #556. Is this still relevant?

PR #692, opened 2022-09-01
fix(mangrove.js): trying to replicate bug

peterMangrove:
#619


peterMangrove:
Hey @jkrivine, i'm trying to replicate the bug #619. I saw that you replicated the bug doing the demo last week. I therefor tried creating a test from the demo script. I'm starting a local node `anvil --fork-url <alchemy-node-Url> --port 8546 ` and the running the test on that local node. But doing this, everything works fine. Do you any easy way of replicating the bug?

Issue #187, opened 2022-02-04
Failing offer bot


dontrolle:
To
* harden our components running in production, and,
* create more realistic noise in our order books

it would be helpful to have a bot, which posts failing offers at regular intervals and reports if they aren't cleaned within a certain timeframe.

dontrolle:
It is part of solving this issue to figure out how to identify and produce offers, which will fail when posted.

This issue currently tracks the work from designing the functionality and workings of the bot, accross implementing it, to deploying on our hosting provider.

The template for bots, here https://github.com/mangrovedao/mangrove/tree/master/packages/bot-template, should give an idea about how to setup the scaffolding for implementing a bot. 

But do look at how one of the bots, which are working in prod (e.g., updategas-bot, cleaning-bot) perform. And although we try to keep `template-bot` updated, there *may* also have been updates to the scaffolding which isn't reflected in `template-bot`. 

Should it be found in the process of working on this issue, that the `template-bot` is of no help, then it should be considered whether it should rather be removed from the codebase.

jkrivine:
Any offer posted without access to liquidity will fail when matched. For instance in the api:
```typescript
let mgv = await Mangrove.connect({ signer: wallet });

//connecting mgv to a market
let market = await mgv.market({ base: "DAI", quote: "USDC" });

// create a simple LP on `market`
let directLP = await mgv.liquidityProvider(market);
//
////Ask on market (promise base (DAI) in exchange of quote (USDC))
//// LP needs to approve Mangrove for base transfer (skipping this part will ensure offers posted by this LP will fail)

// // querying mangrove to know the bounty for posting a new Ask on `market`
let prov = await directLP.computeAskProvision();
tx = await directLP.fundMangrove(prov);
await tx.wait();
//
// //Posting a new Ask (that will fail)
const { id: ofrId } = await directLP.newAsk({ wants: 105, gives: 104 });
```

peterMangrove:
@jkrivine Can you take a look at the bot. Just to see that it does what you think it should do.  Troels and Lasse have already reviewed the actual code, but we would like to be sure, that the actual function of the bot is correct.

espendk:
@peterMangrove This issue is done, right?

peterMangrove:
> @peterMangrove This issue is done, right?

Yes, i'll close the issue

PR #696, opened 2022-09-02
feat(bot): added failing offer bot + bot utils package

peterMangrove:
Added a new bot that creates failing offers, using the example from #187. The bot works just like the maker noise bot and have the same configurations.

Doing this i've added a new package "bot-utils". This gathers all the code that is needed to start a bot and help e.g. post an offer, approve on mangrove etc. All bots have been updated to use this new package. This makes all index files for every bot very simple and easier to implement new bots.

#187


peterMangrove:
@jkrivine  The code has already been reviewed by Troels and Lasse. If you can just take a look the at the bot and if it is in fact doing what you want it to do.

jkrivine (approved):

PR #712, opened 2022-09-13
MGV, _MGV and _mgv to mgv in ctors

lnist:
IMHO MGV and _MGV should be changed to be mixedCase. I am a bit unsure whether _mgv should stay and be applied all over. We seem to use a mix of _ and not for parameter names. I see, e.g., https://ethereum.stackexchange.com/questions/56443/what-does-mean-before-a-variable-name mentions that some do it for all parameters to avoid shadowing. We currently do it when there are collisions but also sometimes when there is not.

I have not changed all tests.


adhusson (approved):
lgtm, let's merge as is or finish updating tests then merge!

PR #715, opened 2022-09-14
Json-deploy

adhusson:
Whenever a script runs, if it calls `outputDeployment()`, it will write its current internal `name=>address` mapping to `addresses/<network>.json`. You need to call `ens.set(name,addr)` after a `new <Contract>` to update the mapping in your script (but you were doing that anyway to be compatible with mangrove.js tests).

Then mangrove.js builds incorporate that information into `addresses.json`.

This is either the final version of address management, or a stopgap until we get a proper `addressProvider`.

Most changes are deleted files in deployments/. The commit 900475fb490f7258549ef96435c583cf3b6a9191 contains what you want to look at.


packages/mangrove-solidity/.gitignore
```diff
@@ -10,4 +10,5 @@ cache
 
 
 # broadcast files, ignore local deploys
-broadcast/*
\ No newline at end of file
+broadcast/*
```
lnist, line 13: Consider ignoring broadcast/**/dry-run/*.json instead, due to  https://github.com/foundry-rs/foundry/issues/3088
Though I see you did not actually change this line :)

packages/mangrove-solidity/copyAddresses.js
```diff
@@ -0,0 +1,15 @@
...
+const path = require("path");
+const shell = require("shelljs");
+shell.config.fatal = true; // throw if a command errors
+
+const sourceDir = path.resolve("./addresses");
+const outDir = path.resolve("./dist/addresses");
+
+shell.cd(sourceDir);
+shell.mkdir("-p", outDir);
+shell.cp("*", outDir);
```
lnist, line 15: Old previously built but now removed addresses files will not be cleaned, but I think that is ok. Just a remark.

packages/mangrove-solidity/script/lib/Deployer.sol
```diff
@@ -4,23 +4,35 @@ import {console, stdJson} from "forge-std/Script.sol";
...
 
-   Do not inherit other deployer scripts, just instantiate them and call their
-   .deploy();
+   Do not inherit other deployer scripts! Just instantiate them and call their
+   .deploy() function;
 */
 abstract contract Deployer is Script2 {
   ToyENS ens; // singleton local ens instance
   ToyENS remoteEns; // out-of-band agreed upon toy ens address
+  mapping(uint => string) chainkeys; // out-of-band agreed upon chain names
```
lnist, line 32: I would camelCase it to chainKeys

packages/mangrove-solidity/script/lib/Deployer.sol
```diff
@@ -43,5 +105,38 @@ abstract contract Deployer is Script2 {
...
       remoteEns.set(names, addrs, isToken);
     }
+
+    // known chain, write deployment file
+    if (bytes(chainkeys[block.chainid]).length != 0) {
+      vm.writeFile(file(), ""); // clear file
+      line("[");
+      for (uint i = 0; i < names.length; i++) {
+        bool end = i + 1 == names.length;
+        line("  {");
```
lnist, line 115: Doesn't foundry's json lib support writing? Its ok we do it manually.

packages/mangrove-solidity/script/lib/Deployer.sol
```diff
@@ -43,5 +105,38 @@ abstract contract Deployer is Script2 {
       vm.broadcast();
       remoteEns.set(names, addrs, isToken);
     }
+
+    // known chain, write deployment file
+    if (bytes(chainkeys[block.chainid]).length != 0) {
+      vm.writeFile(file(), ""); // clear file
+      line("[");
+      for (uint i = 0; i < names.length; i++) {
```
lnist, line 113: can we sort the names so they come alphabetically?

lnist (approved):
Minor comments but otherwise lgtm

packages/mangrove-solidity/.gitignore
adhusson: Yes, with the upcoming change (https://github.com/mangrovedao/mangrove/tree/deploy-reads-addresses) I don't think we'll be pushing broadcasts at all anymore.

packages/mangrove-solidity/copyAddresses.js
adhusson: Good point. Will update.

packages/mangrove-solidity/script/lib/Deployer.sol
adhusson: Will be removed soon anyway :)

packages/mangrove-solidity/script/lib/Deployer.sol
adhusson: I don't think it does but if you see anything that does it please send it my way!

packages/mangrove-solidity/script/lib/Deployer.sol
adhusson: As is, note that the order will respect the existing one found in the read file. There's no builtin ordering, do you see strong reasons to write a routine for that?

packages/mangrove-solidity/script/lib/Deployer.sol
lnist: There was so much debate about file writing I thought they had it - but you are right. JSON is only for reading.

PR #718, opened 2022-09-15
fix(doc): Improve comments for resting order in/out tokens

lnist:


jkrivine (commented):
I am not too found of have taker and maker in out tokens. The initial idea to avoid confusion is to always take maker's viewpoint which is also Mangrove's viewpoint (inbound tokens are those that are brought by the taker, outbound tokens are those that the taker receives).

lnist:
> I am not too found of having both taker and maker `in` `out` tokens. The initial idea to avoid confusion is to always take maker's viewpoint which is also Mangrove's viewpoint (inbound tokens are those that are brought by the taker, outbound tokens are those that the taker receives).

I think that is a good stance, so I've reverted that part. Its probably easier to review in the "files changed" tab, since it is actually much fewer changes than originally.

Note: I would suggest we move the offer-making part to its own method so that we only do the in/out switcheroo once as parameters to that method. I did not do it in this commit as it would make it difficult to see what was happening :)

lnist:
> Also I think it was a bad idea to use a "sell/buy, base quote" terminology in the contract. It is not uniform with the design choice we made to only use inbound outbound rather than base quote. I believe we should change this as well

Sounds reasonable. Are you implying that the `TakerOrder` and `TakerOrderResult` should be updated to reflect that? I think it should be a different PR than this one.

jkrivine:
> > Also I think it was a bad idea to use a "sell/buy, base quote" terminology in the contract. It is not uniform with the design choice we made to only use inbound outbound rather than base quote. I believe we should change this as well
> 
> Sounds reasonable. Are you implying that the `TakerOrder` and `TakerOrderResult` should be updated to reflect that? I think it should be a different PR than this one.

Yes!

jkrivine (changes requested):
Looks good. But another point on this: evm gives opcode for timestamps. @adhusson  was saying this should be a better measure than block numbers (since block/sec fluctuates quite a lot). This should facilitates frontend integration by @johanleroch as well.

PR #725, opened 2022-09-19
fix(restingOrder): blocksToLiveForRestingOrder -> timeToLiveForRestingOrder

lnist:
Change blockToLive to timeToLive and use timestamp which is in seconds.


jkrivine (approved):
all good!

PR #737, opened 2022-09-21
fix(restingOrder): Remove trade terminology from solidity code

lnist:
The trade terminology should be in mangrove.js instead.

* fillWants replaces !selling
* in/out replaces base/quote - mangrove.js updated accordingly
* retry loop to snatch re-posted offers is removed for simplicity
* natspec some


packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -25,7 +25,7 @@ contract MangroveOrder is Forwarder, IOrderLogic {
   constructor(IMangrove mgv, address deployer)
     Forwarder(mgv, new SimpleRouter())
   {
-    setGasreq(30000); // fails < 20K
+    setGasreq(30000); // fails < 30K
```
jkrivine, line 28: no actually fails < 20K is correct but I put 30K to be on the safe side

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -47,23 +47,25 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+      // otherwise, the market order stops when `takerGives` units of `inbound_tkn` have been sold.
+      return res.takerGave >= tko.takerGives;
     }
   }
 
   // `this` contract MUST have approved Mangrove for inbound token transfer
   // `msg.sender` MUST have approved `this` contract for at least the same amount
   // provision for posting a resting order MAY be sent when calling this function
-  // gasLimit of this `tx` MUST be at least `(retryNumber+1)*gasForMarketOrder`
+  // gasLimit of this `tx` MUST at least cover filling the market order.
```
jkrivine, line 68: `gasLimit` should cover for a marketOrder and a newOffer actually
Provision for posting a resting order MUST be sent (there are no other way)

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -72,34 +74,28 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
-    }
+    uint pulled = router().pull(
+      tko.inbound_tkn,
+      msg.sender,
+      tko.takerGives,
+      true
+    );
+    require(pulled == tko.takerGives, "mgvOrder/mo/transferInFail");
+
+    (uint takerGot, uint takerGave, uint bounty, uint fee) = MGV.marketOrder({
```
jkrivine, line 87: `(res.takerGot, res.takerGave, res.bounty, res.fee) = MGV.marketOrder({` is using less stack allocation

packages/mangrove-solidity/src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -12,23 +12,33 @@
 
 pragma solidity >=0.8.0;
 pragma abicoder v2;
-import "mgv_src/IMangrove.sol";
+import {IMangrove} from "mgv_src/IMangrove.sol";
 import {IERC20} from "mgv_src/MgvLib.sol";
 
+///@title Interface for resting orders functionality.
 interface IOrderLogic {
+  ///@notice Information for creating a market order and possibly a resting order (offer).
```
jkrivine, line 20: I am not use of the natspec convention. I think you can use `@notice bah` and `@param blah` for describing the sruct as in:
```
///@notice my struct
///@param x is blah
struct MyStruct {
 uint x;
}
```

packages/mangrove-solidity/test/periphery/MgvOrder.t.sol
```diff
@@ -115,17 +115,15 @@ contract MangroveOrder_Test is MangroveTest {
...
   function test_partial_filled_buy_order_returns_residual() public {
     IOrderLogic.TakerOrder memory buyOrder = IOrderLogic.TakerOrder({
-      base: base,
-      quote: quote,
+      outbound_tkn: base,
+      inbound_tkn: quote,
       partialFillNotAllowed: false,
-      selling: false, //i.e buying
-      wants: 2 ether,
+      fillWants: false,
```
jkrivine, line 121: `fillWants` should be `true` for `buying` I reckon (I buy/ I want, I sell/I give). Same remarks applies for all subsequent taker orders

packages/mangrove.js/src/util/trade.ts
```diff
@@ -443,23 +443,24 @@ class Trade {
...
-        quote: market.quote.address,
+        outbound_tkn: outboundTkn.address,
+        inbound_tkn: inboundTkn.address,
         partialFillNotAllowed: params.partialFillNotAllowed
           ? params.partialFillNotAllowed
           : false,
-        selling: orderType === "sell",
-        wants: wants,
+        fillWants: orderType === "buy",
+        takerWants: wants,
```
jkrivine, line 459: yes this one is correct

jkrivine (changes requested):
a few comments/requests in the code annotation

packages/mangrove-solidity/src/strategies/interfaces/IOrderLogic.sol
lnist: hm, I am not sure. I was relying on:
> A public state variable is equivalent to a function for the purposes of NatSpec.
(from https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#documentation-example )

But `forge doc` does not result in anything either way :/
I'll change it :)

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
lnist: Ah :) Updated.

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
lnist: "at least" :p - but yes, I have updated it.

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
lnist: nifty! done.

packages/mangrove-solidity/test/periphery/MgvOrder.t.sol
lnist: Ah, yes - that was a change prior to our discussion and my understanding. And the reason I want to add another test - they all succeed when flipping the bool so no tests actually care about selling/buying.

jkrivine (approved):

PR #741, opened 2022-09-21
fix(prettier) correct prettier version + code fix

peterMangrove:


lnist:
Is this PR going to master or to the audit branch? It seems to contain audit commits

lnist:
Based on comments on slack I've changed the target to be the audit branch ;)

PR #742, opened 2022-09-22
fix(format) formatted solidity files with forge fmt + as pre-commit

peterMangrove:


packages/mangrove-solidity/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "author": "Mangrove DAO",
   "license": "(AGPL-3.0-only OR BSD-2-Clause)",
+  "type": "module",
```
adhusson, line 6: I'm not sure if this has wide-ranging implications? Doing `require('micromatch')` in `lint-staged.config.js` mais render it unnecessary.

Issue #728, opened 2022-09-19
Add more informations in an `OrderResult` regarding resting orders


johanleroch:
**Is your feature request related to a problem? Please describe.**
When I'm doing a TX, the OrderResult returns me only the details for the immediate purchased part, but not for the posted offer.
If I want to show details to the user after the TX succeed, I have to make a difference between what I asked for (from the front app) and what I received (from mangrovejs).
This leads to potential errors and in addition to displaying bad information to the user.

**Describe the solution you'd like**
it would be great to have the `volume` and `unit price` at which the offer was posted.

PR #751, opened 2022-09-23
feat(restingOrder): Add details about resting order to OrderResult

lnist:
* Remove restingOrderId from MangroveOrder's OrderSummary as its given by NewOwnedOffer event
* Move RawOfferToOffer from semibook to tradeEventManagement
* Generalize event parsing and use it to parse events both from MangroveOrder contract and Mangrove contract, and move most of this logic go tradeEventManagement from trade.
* Change partialFill logic to take fee into account, so it matches MangroveOrder.sol's checkCompleteness function

Fixes https://github.com/mangrovedao/mangrove/issues/728


jkrivine (approved):
looks good to me!

PR #776, opened 2022-09-28
chore(solidity): improved packed/unpacked API

adhusson:
Note: no need to review .post.sol files.


packages/mangrove-solidity/src/IMangrove.sol
```diff
@@ -78,12 +76,12 @@ interface IMangrove {
 
   function best(address outbound_tkn, address inbound_tkn) external view returns (uint);
 
-  function config(address outbound_tkn, address inbound_tkn) external view returns (Global.t, Local.t);
+  function config(address outbound_tkn, address inbound_tkn) external view returns (Structs.GlobalPacked, Structs.LocalPacked);
```
lnist, line 79: I am fine with the naming. Based on your previous comments it seems it could not be, e.g., PackedStructs.Global - which I would have preferred. Correct?
Or maybe even Structs.Packed.Global

packages/mangrove-solidity/src/MgvHasOffers.sol
```diff
@@ -20,20 +20,18 @@ pragma solidity ^0.8.10;
 
 pragma abicoder v2;
 
-import {MgvLib, HasMgvEvents, IMgvMonitor} from "./MgvLib.sol";
+import {MgvLib, HasMgvEvents, IMgvMonitor, Structs} from "./MgvLib.sol";
```
lnist, line 23: I like the new imports :) Are the Structs also to be used from outside mangrove code? Would it then make sense to call it MgvStructs?

packages/mangrove-solidity/src/MgvHasOffers.sol
```diff
@@ -20,20 +20,18 @@ pragma solidity ^0.8.10;
...
-import {Offer, OfferDetail, Global, Local} from "mgv_src/preprocessed/MgvPack.post.sol";
-import {OfferStruct, OfferDetailStruct, GlobalStruct, LocalStruct} from "mgv_src/preprocessed/MgvStructs.post.sol";
 
 /* `MgvHasOffers` contains the state variables and functions common to both market-maker operations and market-taker operations. Mostly: storing offers, removing them, updating market makers' provisions. */
 contract MgvHasOffers is MgvRoot {
   /* # State variables */
   /* Given a `outbound_tkn`,`inbound_tkn` pair, the mappings `offers` and `offerDetails` associate two 256 bits words to each offer id. Those words encode information detailed in [`structs.js`](#structs.js).
 
-     The mappings are `outbound_tkn => inbound_tkn => offerId => Offer.t|OfferDetail.t`.
+     The mappings are `outbound_tkn => inbound_tkn => offerId => Structs.OfferPacked|Structs.OfferDetailPacked`.
```
lnist, line 31: I have looked over these changes, but as I assume they are made by search and replace I haven't looked very closely at these mechanical changes.

packages/mangrove-solidity/src/MgvOfferMaking.sol
```diff
@@ -287,7 +286,7 @@ contract MgvOfferMaking is MgvHasOffers {
             || offerDetail.offer_gasbase() != ofp.local.offer_gasbase()
         ) {
           uint offer_gasbase = ofp.local.offer_gasbase();
-          offerDetails[ofp.outbound_tkn][ofp.inbound_tkn][ofp.id] = OfferDetail.pack({
+          offerDetails[ofp.outbound_tkn][ofp.inbound_tkn][ofp.id] = Structs.OfferDetail.pack({
```
lnist, line 289: I thought it would be Structs.OfferDetail(Packed|Unpacked) - or is the one without any postfix the unpacked one? For global its Structs.GlobalUnpacked.

lnist (commented):
I assume you are on top of making this easiest to review for auditors :) If you think its fine, then lets go, otherwise we could consider making some sed commands for the mechanical changes and split into multiple commits. I'll let you decide.

adhusson:
> I assume you are on top of making this easiest to review for auditors :) If you think its fine, then lets go, otherwise we could consider making some sed commands for the mechanical changes and split into multiple commits. I'll let you decide.

I'm interested in having more details on this :)

packages/mangrove-solidity/src/MgvHasOffers.sol
adhusson: Yes, the goal of MgvLib + its dependencies is that people can import the Mangrove API easily. Hmm since they're imported from MgvLib, now I wonder if I should add an `Mgv` prefix... or remove the `Mgv` prefix from the other imports. wdyt?

packages/mangrove-solidity/src/MgvOfferMaking.sol
adhusson: `Structs.OfferDetail` is a namespace, so you can park functions in there. `Structs.OfferDetailPacked` is a type alias, and `Structs.OfferDetailUnpacked` is as struct name. Not sure I understand your question, let me know if that clears things up.

packages/mangrove-solidity/src/IMangrove.sol
lnist: I think those are fine arguments, so lets keep it.
Wrt functions I would park "pack" with the unpacked and "unpacked" with the packed. But since we don't do it, lets not discuss further :)

packages/mangrove-solidity/src/MgvHasOffers.sol
lnist: To keep it consistent would be to add Mgv. I am a bit split since redundant mgv is redundant but on the other hand a tyupe name "Lib" "HasEvents" can quickly clash with types in other libraries that strats may be using. So safe change is to add Mgv to struct.

packages/mangrove-solidity/src/MgvOfferMaking.sol
lnist: I was confused about types vs functions. This made it clear, thanks.

lnist:
> otherwise

You could split this PR in multiple commits which you merge to master. 1) the actual code change to generator, 2) the generated files which do not need review. 3) the mechanical search and replace .
The last commit could then be reviewed by auditors by standing on the previous commit and executing a number of commands in a shell that do nothing but string replacement.

It did not take _that_ long to review, so maybe it is a waste of our time, but maybe they don't trust the "manual" search+replace and have to go over all of them - or they could make the reverse change themselves :)

adhusson:
Script to go from 4019c32601de75f2d006cc16419d919f7d1f0dbf to after this PR: (will help auditors compare for `src/*.sol`) (uses the rool [ruplacer](https://github.com/your-tools/ruplacer)):

```bash
ruplacer ', P,' ', MgvStructs,' --go

ruplacer 'P}' 'MgvStructs}' --go

ruplacer 'Pack.post.sol" as P' 'Structs.post.sol" as MgvStructs' --go

ruplacer 'P\.(.+?)\.t' 'MgvStructs.${1}Packed' --go

ruplacer 'P\.(.+?)Struct' 'MgvStructs.${1}Unpacked' --go

ruplacer 'MgvLib as ML' 'MgvLib' --go

ruplacer 'ML' 'MgvLib' --go

ruplacer cancelations cancellations --go

ruplacer 'offer book' 'order book' --go

ruplacer availale available --go

FOUNDRY_FMT_TAB_WIDTH=2 FOUNDRY_FMT_INT_TYPES="short" FOUNDRY_FMT_NUMBER_UNDERSCORE="thousands" forge fmt

ruplacer 'P\.(.+?)\.pack' 'MgvStructs.${1}.pack' --go

FOUNDRY_FMT_TAB_WIDTH=2 FOUNDRY_FMT_INT_TYPES="short" FOUNDRY_FMT_NUMBER_UNDERSCORE="thousands" forge fmt

git apply << EOM
diff --git a/packages/mangrove-solidity/src/MgvHasOffers.sol b/packages/mangrove-solidity/src/MgvHasOffers.sol
index 8e4f1e03..7d149f46 100644
--- a/packages/mangrove-solidity/src/MgvHasOffers.sol
+++ b/packages/mangrove-solidity/src/MgvHasOffers.sol
@@ -123,14 +123,15 @@ contract MgvHasOffers is MgvRoot {
     MgvStructs.LocalPacked local
   ) internal returns (MgvStructs.LocalPacked) {
     unchecked {
+      mapping(uint => MgvStructs.OfferPacked) storage semiBook = offers[outbound_tkn][inbound_tkn];
       if (betterId != 0) {
-        offers[outbound_tkn][inbound_tkn][betterId] = offers[outbound_tkn][inbound_tkn][betterId].next(worseId);
+        semiBook[betterId] = semiBook[betterId].next(worseId);
       } else {
         local = local.best(worseId);
       }
 
       if (worseId != 0) {
-        offers[outbound_tkn][inbound_tkn][worseId] = offers[outbound_tkn][inbound_tkn][worseId].prev(betterId);
+        semiBook[worseId] = semiBook[worseId].prev(betterId);
       }
 
       return local;
EOM
```

lnist (approved):

PR #777, opened 2022-09-29
Invariant annotations + refactoring

jkrivine:
Need to check
- moving the transfer of market order residual earlier
- always using router to transfer ERC to msg.sender's reserve (in case the reserve is not EOA)
- invariants in the comments do hold and are clear

### DIFF
--- a/packages/mangrove-solidity/src/periphery/MangroveOrder.sol
+++ b/packages/mangrove-solidity/src/periphery/MangroveOrder.sol
@@ -56,7 +56,9 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     return super.__lastLook__(order);
   }
 
-  ///@notice checks whether the order is completely filled
+  ///@notice compares a taker order with a market order results and checks whether the order was entirely filled
+  ///@param tko the taker order
+  ///@param res the market order result
   function checkCompleteness(TakerOrder calldata tko, TakerOrderResult memory res) internal pure returns (bool) {
     // The order can be incomplete if the price becomes too high or the end of the book is reached.
     if (tko.fillWants) {
@@ -69,12 +71,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
   }
 
   // `this` contract MUST have approved Mangrove for inbound token transfer
-  // `msg.sender` MUST have approved `this` contract for at least the same amount
-  // provision for posting a resting order MUST be sent when calling this function
-  // gasLimit of this `tx` MUST at cover filling the market order and a new offer for resting orders.
-  // msg.value SHOULD contain enough native token to cover for the resting order provision
-  // msg.value MUST be 0 if `!restingOrder` otherwise transferred WEIs are burnt.
-
+  // `msg.sender` MUST have approved `this` contract's router for at least the same amount
+  // provision for posting a resting order MUST be sent when calling this function  
   function take(TakerOrder calldata tko) external payable returns (TakerOrderResult memory res) {
     // pulling directly from msg.sender would require caller to approve `this` in addition to the `this.router()`
     // so pulling funds from taker's reserve (note this can be the taker's wallet depending on the router)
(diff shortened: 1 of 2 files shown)


packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -93,31 +91,42 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+    // 2. `this` contract inbound and outbound token balances are back to their initial value
+    // 3. `this` contract's WEI balance is credited of `msg.value + res.bounty`
 
     if (tko.restingOrder && !isComplete) {
-      // When posting a resting order the taker becomes a maker, and `inbound_tkn` for an offer is what the maker receives,
-      // so the offer for this resting order must have `inbound_tkn` set to `outbound_tkn` such that the taker
-      // receives these when the offer is taken, and the offer's `outbound_tkn` becomes `inbound_tkn`.
-      postRestingOrder({tko: tko, outbound_tkn: tko.inbound_tkn, inbound_tkn: tko.outbound_tkn, res: res});
+      // When posting a resting order `msg.sender` becomes a maker. 
+      // For maker orders outbound tokens are what makers send. Here `msg.sender` wants to send `tko.inbound`.
```
lnist, line 109: tko.inbound_tkn

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -93,31 +91,42 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+      // The offer list on which this contract must post taker's resting order is thus `(tko.inbound_tkn, tko.outbound_tkn)`
```
lnist, line 110: taker=>msg.sender to be consistent with above?

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -93,31 +91,42 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+      // the call below will fill the memory data `res`.
+      postRestingOrder({
+        tko: tko, 
+        outbound_tkn: tko.inbound_tkn, 
+        inbound_tkn: tko.outbound_tkn, 
+        res: res, 
+        fund: msg.value + res.bounty // using bounty as additional funds to provision the resting order
+      });
+      // at this point the following invariants hold:
+      // 1. `OrderSummary` was logged and either (2) or (2') hold.
```
lnist, line 120: also add that the msg.sender's native token wrt this contract is now back to the outset (0?)?

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -143,7 +152,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     TakerOrder calldata tko,
     IERC20 inbound_tkn,
     IERC20 outbound_tkn,
-    TakerOrderResult memory res
+    TakerOrderResult memory res,
+    uint fund
```
lnist, line 158: comment on line 160 should be updated now that fund is a parameter which replaces msg.value+res.bounty

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -187,8 +197,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
         "mgvOrder/mo/transferInFail"
       );
       // msg.value is no longer needed so sending it back to msg.sender along with possible collected bounty
```
lnist, line 200: msg.value=>`fund`

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -187,8 +197,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
         "mgvOrder/mo/transferInFail"
```
lnist, line 198: We have already pushed the funds back to msg.sender on line 99 so I am a bit confused now. Can you remind me what is the difference between using the router and transferLib?

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
```diff
@@ -187,8 +197,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
       );
       // msg.value is no longer needed so sending it back to msg.sender along with possible collected bounty
-      if (msg.value + res.bounty > 0) {
-        (bool noRevert,) = msg.sender.call{value: msg.value + res.bounty}("");
+      if (fund > 0) {
+        (bool noRevert,) = msg.sender.call{value: fund}("");
```
lnist, line 202: consider whether it would be better to have this method return fund or 0 depending on whether it should be refunded, and then let the main method return it so that all refunding happens there in a single call.

lnist (changes requested):

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
jkrivine: good catch this line should be removed

packages/mangrove-solidity/src/periphery/MangroveOrder.sol
jkrivine: good point

PR #785, opened 2022-10-01
feat(solidity): optimize struct slot computation across Mangrove

adhusson:
This PR reinstates & extends @lnist 's partially evaluated storage pointer optimizations.

Questions:
1. Is it correct?
2. Is it worth it?

The gas usage and differences wrt master are (for tests in `Gas.t.sol`):
* test_market_order_1
  * Gas used: 155558 **(-443)**
* test_market_order_8
  * Gas used: 477017 **(-5391)**
* test_new_offer
  * Gas used: 81628 **(-690)**
* test_partial_take_offer
  * Gas used: 155424 **(-11)**
* test_take_offer
  * Gas used: 155413 **(-11)**
* test_update_full_offer
  * Gas used: 36725 **(-229)**
* test_update_min_move_0_offer
  * Gas used: 24423 **(-205)**
* test_update_min_move_3_offer
  * Gas used: 21399 **(-1206)**
* test_update_min_move_6_offer
  * Gas used: 22953 **(-1701)**

┆Issue is synchronized with this [Monday item](https://mangrovedao.monday.com/boards/3894180513/pulses/3901913077) by [Unito](https://www.unito.io)

### DIFF
--- a/packages/mangrove-solidity/src/MgvOfferMaking.sol
+++ b/packages/mangrove-solidity/src/MgvOfferMaking.sol
@@ -22,6 +22,7 @@ pragma abicoder v2;
 
 import {IMaker, HasMgvEvents, MgvStructs} from "./MgvLib.sol";
 import {MgvHasOffers} from "./MgvHasOffers.sol";
+import "forge-std/console.sol";
 
 /* `MgvOfferMaking` contains market-making-related functions. */
 contract MgvOfferMaking is MgvHasOffers {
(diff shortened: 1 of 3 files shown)


packages/mangrove-solidity/src/MgvOfferMaking.sol
```diff
@@ -46,6 +46,8 @@ contract MgvOfferMaking is MgvHasOffers {
     MgvStructs.LocalPacked local;
     // used on update only
     MgvStructs.OfferPacked oldOffer;
+    bytes32 semibook;
```
lnist, line 49: This optimization comes at the cost of increased memory usage in this structure - do we have a lot of these or only a few at at time? In the latter case I guess it is fine, but otherwise we have to make a conscious decision on whether the memory consumption can grow too much with many offers, such that gas optimizations are lost.

lnist (commented):
I would have liked solidity to support this wince it is quite a low level way to reference the data.
Additionally, @jkrivine mentioned we may want to postpone this change and have it on a gas-optimization/mgv-v1.1 branch for later auditing. You two figure that out :)

packages/mangrove-solidity/src/MgvOfferMaking.sol
adhusson: Looks like my reply was never posted! There's only one at a time :)

adhusson:
outdated -- merge in mangrove-core

PR #789, opened 2022-10-03
Feature/improve deployers

jkrivine:
To check:
- mangrove market activation scripts (6bfcca91 & 706a7c260b)
- erc20 deployer (9e1370c78f)


adhusson:
putting this on hold until new repo

packages/mangrove-solidity/script/ActivateSemibook.s.sol
```diff
@@ -64,6 +64,8 @@ contract ActivateSemibook is Test2, Deployer {
     (MgvStructs.GlobalPacked global,) = mgv.config(address(0), address(0));
     uint outbound_decimals = ERC20(outbound_tkn).decimals();
     uint density = (COVER_FACTOR * global.gasprice() * 10 ** outbound_decimals) / outbound_in_gwei;
+    // density should be at least 1 wei delivered per gas units
```
adhusson, line 67: Why?

packages/mangrove-solidity/script/ERC20Deployer.sol
```diff
@@ -0,0 +1,35 @@
...
+    require(uint8(dec) == dec, "Decimals overflow");
+    broadcast();
+    TestToken token = new TestToken({
+      admin: msg.sender,
+      name: vm.envString("NAME"),
+      symbol: symbol,
+      _decimals: uint8(dec)
+    });
+    // smoke test
+    require(token.decimals() == dec, "Smoke test failed");
```
adhusson, line 31: I don't underestand this test, assuming TestToken is not buggy, it is statically provable that `token.decimals() == dec`.

packages/mangrove.js/examples/demo-GNT.js
```diff
@@ -0,0 +1,21 @@
+// this script assumes dotenv package is installed `(npm install dotenv --save)`
```
adhusson, line 1: Should this file be ported to mangrove-core?

packages/mangrove-solidity/script/ActivateSemibook.s.sol
jkrivine: because of underflow. Normally you should have something like 0.6 wei per gas unit. Allowing density to be 0 paves the way for free grieffing

packages/mangrove-solidity/script/ERC20Deployer.sol
jkrivine: The smoke test just call some function to make sure the contract is indeed deployed. Its not a test of `decimals()`

packages/mangrove.js/examples/demo-GNT.js
jkrivine: hum no

packages/mangrove-solidity/script/ERC20Deployer.sol
adhusson: I don't think it can ever happen that `new Contract` does not create the contract. Unless forge itself has a very grave bug. Also note that this will only do the test during the script execution phase, not during the simulation, nor during the broadcast.

packages/mangrove-solidity/script/ActivateSemibook.s.sol
adhusson: Is there an important threshold between 0 and nonzero, as in will it deter a class of attacks that would have been profitable at 0? Otherwise if it's just 'nice to have' I'd rather not have it, if it gives a false sense of security.

PR #802, opened 2022-10-06
refactor(mangrove.js): Rename penalty to bounty in Market.Summary

espendk:
It's a bounty from the taker's (who's buying or selling) perspective, so makes more sense to call it that.


lnist (approved):
Agree!

PR #803, opened 2022-10-06
refactor(mangrove.js): don't reexport addresses/abis

adhusson:


packages/mangrove.js/src/constants/addresses.json
```diff
@@ -15,13 +14,6 @@
...
     "DAI": "0xc87385b5e62099f92d490750fcd6c901a524bbca",
     "USDC": "0xF61Cffd6071a8DB7cD5E8DF1D3A5450D9903cF1c",
-    "WETH": "0x63e537a69b3f5b03f4f46c5765c82861bd874b6e",
-    "MangroveOrder": "0x1fA0d582B8aA298f634cdCDCCcB7972E913b2D8f",
-    "AaveProvider": "0x5343b5bA672Ae99d627A1C87866b8E53F47Db2E6",
-    "Mangrove": "0xF3e339d8a0B989114412fa157Cc846ebaf4BCbd8",
-    "MgvReader": "0xfAB31d37f8DF5bff07Bb3c16B33416eCd4Aab76F",
-    "MgvCleaner": "0xEb05Ace3574B0a6f4696c5CcD09e730d6d5ED3b0",
-    "MgvOracle": "0xd38c02425da847584eeDA72387DAAA2E8f3b90c8",
-    "MangroveOrderEnriched": "0x9A643978A0A50459d6159bc87f97B309E539083b"
```
lnist, line 25: are these not used by the `market.mgv.orderContract` statements or by the ui?

lnist (commented):
I am not entirely sure why we had all the addresses prior - weren't they used by ui or by js code? If not then I love it!

packages/mangrove.js/src/constants/addresses.json
adhusson: They are used, but the changes to constants.ts should ensure they are still available. Lmk what you think.

adhusson:
Hmm it looks like I may have removed too many abis.

packages/mangrove.js/src/constants/addresses.json
lnist: ah, of course!

lnist (approved):

adhusson:
Ok, I had removed the correct amount of ABIs (all of them) but forgot to update updatebot

vd1:
what is update bot doing?

PR #831, opened 2022-10-25
CI: Don't fail test report steps

espendk:
The upload of the test report would fail if any tests had failed. This had the annoying consequence that when clicking a failed job in GitHub Actions, it would scroll to the test report step instead of to the logs of the failing test itself.

Now the test report steps succeed even if any tests failed. The job will still fail in that case, so the flow of the workflow is unchanged.


espendk:
Here's a run that demonstrates this working (the PR checks are useless since they run on `master`'s version of the workflow): https://github.com/mangrovedao/mangrove/actions/runs/3319471957/jobs/5484685804

peterMangrove (approved):

PR #855, opened 2022-11-10
fix(trade): use correct gaslimit when sniping

peterMangrove:
When trying to snipe an offer without setting a gasLimit, it would go through the order book to find the correct gasLimit. This is every time consuming. Now we instead use the offers gasReq.

┆Issue is synchronized with this [Monday item](https://mangrovedao.monday.com/boards/3894180513/pulses/3901897216) by [Unito](https://www.unito.io)


espendk:
Is this still relevant @peterMangrove or can we close it?

If it's still relevant, I guess it should be updated to target `develop` instead of `master`

lnist:
Yes I believe this is still an issue.

espendk:
Ok, is it ready for review?
I don't see any added/changed tests, but maybe this code is sufficiently covered by existing tests?

lnist:
I now believe this fix is wrong. A snipe's gas usage is always limited by the gasreq of the offer, so no need to pass it in. In that case we can just pass in a very big number (above 2^24-1) . I'm including the fix in another patch.

PR #861, opened 2022-11-14
feat(trade): added fill or kill option to buy and sell

peterMangrove:


packages/mangrove.js/src/util/trade.ts
```diff
@@ -500,6 +504,99 @@ class Trade {
     return result;
   }
 
+  async fok(
```
lnist, line 507: I suggest extending "restingOrder" to have the fok parameter available and renaming it to "mangroveOrder".

packages/mangrove.js/src/util/trade.ts
```diff
@@ -220,7 +222,9 @@ class Trade {
   ): Promise<Market.OrderResult> {
     const { gives, wants, wantsWithoutSlippage, fillWants } =
       this.getParamsForSell(params, market.base, market.quote);
-    if ("restingOrder" in params && params.restingOrder) {
+    if ("fillOrKill" in params && params.fillOrKill === "yes") {
```
lnist, line 225: I would just make it a boolean like restingOrder and then let a later refactoring make it cleaner

packages/mangrove.js/src/util/trade.ts
peterMangrove: Yes, makes sense

packages/mangrove.js/src/util/trade.ts
```diff
@@ -500,6 +504,99 @@ class Trade {
...
+    overrides_.gasLimit = overrides_.gasLimit
+      ? overrides_.gasLimit
+      : await market.estimateGas(orderType, wants);
+
+    const [outboundTkn, inboundTkn] =
+      orderType === "buy"
+        ? [market.base, market.quote]
+        : [market.quote, market.base];
+
+    const response = await market.mgv.orderContract.take(
```
jkrivine, line 533: this is the old interface of MangroveOrder. I think we should use the new one because this one will be obsolete soon

peterMangrove: I would like to keep the switch from the old interface to the new one. In a separate PR.

lnist (approved):

PR #870, opened 2022-11-17
feat(restingOrder): Upgrade to new mangrove-core supporting expiry as a timestamp

lnist:
… a timestamp


packages/mangrove.js/src/market.ts
```diff
@@ -758,30 +757,37 @@ class Market {
...
 
-      txPromise.then((resp) => {
-        // Warning: if the tx nor any with the same nonce is ever mined,
-        // the `once` and block callbacks will never be triggered and you will memory leak by queuing tasks.
-        txHashDeferred.resolve(resp.hash);
-        resp
-          .wait(1)
-          .then((recp) => {
-            this.afterBlock(recp.blockNumber, async () => {
+      txPromise
```
adhusson, line 760: the prettification makes it difficult to understand what has changed

packages/mangrove.js/src/market.ts
```diff
@@ -758,30 +757,37 @@ class Market {
...
             });
-          })
-          .catch((e) => {
-            this.unsubscribe(cb);
-            ko({ revert: true, exception: e });
-          });
-      });
+        })
+        .catch((e) => {
+          txHashDeferred.reject(e);
```
adhusson, line 787: yes seems good! agreed that this is just cleaning up our objects and there is no behavior change, right?

packages/mangrove.js/src/offerLogic.ts
```diff
@@ -154,6 +154,7 @@ class OfferLogic {
   ): Promise<Big> {
     const rawBalance = await this.contract.tokenBalance(
       this.mgv.getAddress(tokenName),
+      await this.mgv._signer.getAddress(),
```
adhusson, line 157: not sure what this is for

packages/mangrove.js/test/unit/trade.unit.test.ts
```diff
@@ -52,7 +52,9 @@ describe("Trade unit tests suite", () => {
       const expectedGivesWithoutSlippage = Big(params.volume).mul(price);
       assert.equal(result.wants.eq(BigNumber.from(params.volume)), true);
       assert.equal(
-        result.givesWithoutSlippage.eq(expectedGivesWithoutSlippage),
+        BigNumber.from(expectedGivesWithoutSlippage.toFixed(0)).eq(
```
adhusson, line 55: why did the equality test change from
`a.eq(b)` to `BigNumber.from(b.toFixed(0)).eq(a'))` ?

packages/mangrove.js/src/offerLogic.ts
lnist: the tokenBalance now takes a maker as a parameter.

packages/mangrove.js/src/market.ts
lnist: yeah - I use kdiff3 to get something meaningful. See slack :)

packages/mangrove.js/src/offerLogic.ts
jkrivine: I think tokenBalance should have an optional `maker` parameter. If none are provided then `this.mgv._signer.getAddress()` should be used

jkrivine (changes requested):
Looks good to me except maybe a minor change in `tokenBalance`

packages/mangrove.js/src/market.ts
```diff
@@ -758,30 +755,43 @@ class Market {
...
+                    revert: false,
+                    exception: "tx mined but filter never returned true",
+                  });
+                }
+              })
+                .then(() => {
+                  // ignored
+                })
+                .catch((e) => {
+                  throw e;
```
adhusson, line 781: I think it's actually fine without it... sorry!

lnist: removed :)

jkrivine (approved):

Issue #871, opened 2022-11-18
Add automatic price calculation in `market.{buy|sell}`


espendk:
**Is your feature request related to a problem? Please describe.**
`market.{buy|sell}` used to allow `price` to be set to `null` which meant "accept any price" (= simulate a market order). This is dangerous, as such tx's are likely to be sandwich attacked:

1. Attacker borrows liquidity to front-run the victim
2. Attacker buys/sells the requested volume on Mangrove and inserts an offer at the top of the book with the requested volume and a price that is just a tiny bit better than the remainder on the book
3. Victim's tx is executed and buys the attacker's offer
4. Attacker repays the loan and pockets the difference between the borrowed amount and what the victim paid.

While we cannot prevent users from sending such orders, it should not be encouraged in the API. Therefore this feature was removed in commit 1a0d9de4fa69b95b068d4365b3e427650b967745.

However, having automatic calculation of the price is useful, so a replacement feature should be implemented.

**Describe the solution you'd like**
When `price=null`, estimate the avg price for the order based on the cache and then use that price.

**Describe alternatives you've considered**
Best offer price + 5%. However, that seems like a non-obvious behavior to me. You'll ask to buy a certain quantity and behind the scenes the SDK chooses an an arbitrary price.

**Additional context**
N/A

adhusson:
My take on this is that exploratory use of the ~API~SDK (eg in a REPL) can often lead to trying out `mgv.marketOrder({wants: amount, price:???})` and the exact price is not important or difficult to figure out (because the user doesn't know the units yet, or because the tokens are fake, prices are arbitrary, etc).

The natural solution is to say: no price means pure market order, but indeed it is too dangerous to leave that as the default.

Uglier but maybe good enough:
* if `chainId=31337` or `network=localhost`, pure market order
* error otherwise
wdyt?

espendk:
I think having the behaviour differ locally and in production is a recipe for disaster 😅

I think we have the necessary price estimation functions already, so I think it's just a matter of invoking that when `price=null`. But it's still not a safe practice, so I'm actually not that keen on making such automagic too easy in the SDK.

lnist:
I would argue the SDK user can invoke the simulate functions themselves and use the result. Can we close this @espendk ?

espendk:
> I would argue the SDK user can invoke the simulate functions themselves and use the result. Can we close this @espendk ?

Agreed!

PR #872, opened 2022-11-18
feat!(market): Remove dangerous behaviour when price is null

espendk:
`market.{buy|sell}` allowed `price` to be set to `null` which meant "accept any price" (= simulate a market order). This is dangerous, as such tx's are likely to be sandwich attacked:

1. Attacker borrows liquidity to front-run the victim
2. Attacker buys/sells the requested volume on Mangrove and inserts an offer at the top of the book with the requested volume and a price that is just a tiny bit better than the remainder on the book
3. Victim's tx is executed and buys the attacker's offer
4. Attacker repays the loan and pockets the difference between the borrowed amount and what the victim paid.

While we cannot prevent users from sending such orders, it should not be encouraged in the API. Therefore this commit removes the feature.

At a later stage we can add safer semantics for price=null. Eg to estimate the avg price for the order based on the cache and then use that price. See issue #871.


adhusson:
Good catch 👍

PR #876, opened 2022-11-21
fix(test): Make proxy between tests and anvil to allow closure

lnist:
Integration tests run against an anvil instance. We use snapshots and in `beforeEach` we revert to a previous, good, snapshot before starting each test.

However, the prior test can have outstanding RPC requests towards the anvil instance and these can complete _after_ the revert, causing changes to anvil state.

We now proxy all requests through a proxy and close it for further requests and wait for outstanding requests before reverting.

This fixes sporadic test failures like
https://github.com/mangrovedao/mangrove-ts/actions/runs/3410881559/jobs/5674398420
and
https://github.com/mangrovedao/mangrove-ts/actions/runs/3411610454/jobs/5676045661
where revert does not do its job.

An alternative fix would be to figure out why we have outstanding requests - but they should not make random tests fail.
Also, we could consider a feature in anvil where it can close connections/switch ports.
Spinning up a new anvil instance defeats the purpose of fast execution of tests.

### DIFF
--- a/packages/mangrove.js/src/util/node.ts
+++ b/packages/mangrove.js/src/util/node.ts
@@ -10,7 +10,7 @@
 const childProcess = require("child_process");
 const path = require("path");
 const fs = require("fs");
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
 import * as eth from "../eth";
 import { Mangrove } from "..";
 import * as ToyENS from "./ToyENSCode";
@@ -294,7 +294,7 @@ const deploy = async (params: {
 const connect = async (params: {
   spawn: boolean;
   deploy: boolean;
-  url: boolean;
+  url: string;
   provider: JsonRpcProvider;
   stateCache: boolean;
   targetContract: string;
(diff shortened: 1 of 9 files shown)


packages/mangrove.js/src/util/node.ts
```diff
@@ -330,15 +330,27 @@ const connect = async (params: {
     params,
     snapshot: async () => {
       lastSnapshotId = await params.provider.send("evm_snapshot", []);
-      snapshotBlockNumber = await params.provider.getBlockNumber();
+
+      const bn = await params.provider.perform("getBlockNumber", {});
```
adhusson: why `getBlockNumber` -> `perform("getBlockNumber)`?

packages/mangrove.js/src/util/node.ts
```diff
@@ -330,15 +330,27 @@ const connect = async (params: {
...
       await params.provider.send("evm_revert", [snapshotId]);
-      const blockNumberAfterRevert = await params.provider.getBlockNumber();
+
+      const bn = await params.provider.perform("getBlockNumber", {});
+
+      const blockNumberAfterRevert = BigNumber.from(bn).toNumber();
       if (blockNumberAfterRevert != snapshotBlockNumber) {
         throw Error(
-          `evm_revert did not revert to expected block number ${snapshotBlockNumber} but to ${blockNumberAfterRevert}. Snapshots are deleted when reverting - did you take a new snapshot after the last revert?`
+          `evm_revert aadid not revert to expected block number ${snapshotBlockNumber} but to ${blockNumberAfterRevert}. Snapshots are deleted when reverting - did you take a new snapshot after the last revert?`
```
adhusson: I think some vim commands have sneaked up here :)

adhusson (approved):
Great catch. Also super clean&short implementation of something that seemed daunting when I read the description.

packages/mangrove.js/src/util/node.ts
lnist: Sorry - note.ts should almost not have been updated. Fixed.

packages/mangrove.js/src/util/test/mochaHooks.ts
```diff
@@ -60,7 +64,51 @@ export const mochaHooks = {
...
+
+    // Create a new proxy for a new port (in case an outstanding async operation for a previous test sends a request)
+    const newProxy = {
+      cancelAll: false,
+      proxyServer: null,
+    };
+    currentProxyPort++;
+    newProxy.proxyServer = new ProxyServer({
+      upstream: async function () {
+        return "127.0.0.1:8545";
```
adhusson, line 91: Shouldn't this be
```suggestion
        return `${serverParams.host}:${serverParams.port}`;
```

adhusson (changes requested):
Small change for port consistency

adhusson (approved):

PR #877, opened 2022-11-21
docs(market): Update comment missed in @ffd5f6f

espendk:


packages/mangrove.js/src/market.ts
```diff
@@ -513,7 +513,7 @@ class Market {
    *
    * @example
    * ```
-   * const market = await mgv.market({base:"USDC",quote:"DAI"}
+   * const market = await mgv.market({base:"USDC",quote:"DAI"})
```
vd1, line 516: should there be a semi-column here line 516? and a closing parenthesis line 487?

PR #931, opened 2023-01-02
feat: add Mangrove.openMarkets[Data]

adhusson:
* add Multicall2 to typechain generation
* let typechain generate from multiple places
* include 'cashness' concept in Mangrove, add constants/tokenCashness.json
* openMarkets returns Market objects (intialized or not)
* openMarketsData returns lower-level data, but takes fewer RPC calls


packages/mangrove.js/src/mangrove.ts
```diff
@@ -50,6 +51,13 @@ namespace Mangrove {
     gasmax: number;
     dead: boolean;
   };
+
+  export type openMarketInfo = {
```
lnist, line 55: We normally use PascalCasing for types (except for devNode.ts :D (maybe that should be changed also? :) )

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
       setAddress(name, address, decimals);
     }
   }
+
+  /**
+   * Returns open markets data according to the local mangrove reader.
```
lnist, line 492: What is meant by "local" in this sentence?

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+   * @param from: start at market i
+   * @param maxLen: max number of markets returned
+   * @param configs: fetch markets config information
```
lnist, line 495:
```suggestion
   * @param configs: fetch market's config information. Default true.
```

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
       setAddress(name, address, decimals);
     }
   }
+
+  /**
+   * Returns open markets data according to the local mangrove reader.
+   * @param from: start at market i
```
lnist, line 493:
```suggestion
   * @param from: start at market `from`. Default 0.
```

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+  async openMarketsData(
+    params: {
+      from?: number;
+      maxLen?: number | ethers.BigNumber;
+      configs?: boolean;
+      tokenInfos?: boolean;
+    } = {}
+  ): Promise<Mangrove.openMarketInfo[]> {
+    // set default params
+    params.from = "from" in params ? params.from : 0;
```
lnist, line 508: This can be simplified to params.from ??= 0
Similar for other defaulting.

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
       setAddress(name, address, decimals);
     }
   }
+
+  /**
+   * Returns open markets data according to the local mangrove reader.
+   * @param from: start at market i
+   * @param maxLen: max number of markets returned
```
lnist, line 494:
```suggestion
   * @param maxLen: max number of markets returned. Default all.
```

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+              data[quote].configs[base],
+              data[quote].decimals
+            )
+          : undefined,
+      };
+    });
+  }
+
+  /**
+   * Returns open markets according to the local mangrove reader. Will internally update Mangrove token information.
```
lnist, line 604: Also update this comment if you update the one for the method above.

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+   */
+  async openMarkets(
+    params: {
+      from?: number;
+      maxLen?: number;
+      noInit?: boolean;
+      bookOptions?: Market.BookOptions;
+    } = {}
+  ): Promise<Market[]> {
+    const noInit = "noInit" in params ? params.noInit : false;
```
lnist, line 619: same simplification as above. The `"noInit" in params` construct is mostly used to satisfy typescript in cases where we don't know if the property is there based on types.

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+    delete params.noInit;
```
lnist, line 620: ?!

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+          quote: quote.symbol,
+          bookOptions: bookOptions,
+          noInit: noInit,
+        });
+      })
+    );
+  }
+
+  // relative cashness of a token will determine which is base & which is quote
+  // lower cashness is base, higher cashness is quote, tiebreaker is <
```
lnist, line 652:
```suggestion
  // lower cashness is base, higher cashness is quote, tiebreaker is lexicographic ordering of symbol string
```

lnist (changes requested):

packages/mangrove.js/src/mangrove.ts
adhusson: I meant the local chain but that's a terrible way of phrasing things

packages/mangrove.js/src/mangrove.ts
adhusson: thanks, done! not found of having a different syntax for defaulting optional/non-optional fields, but I like the shorthand

packages/mangrove.js/src/mangrove.ts
adhusson: I'm passing params to `openMarketsData` so I thought it better to clean the object. wdyt?

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,190 @@ class Mangrove {
...
+      maxLen?: number;
+      noInit?: boolean;
+      bookOptions?: Market.BookOptions;
+    } = {}
+  ): Promise<Market[]> {
+    const noInit = "noInit" in params ? params.noInit : false;
+    delete params.noInit;
+    const bookOptions =
+      "bookOptions" in params ? params.bookOptions : undefined;
+    delete params.bookOptions;
```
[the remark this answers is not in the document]
adhusson: same as above

packages/mangrove.js/src/mangrove.ts
adhusson: 😅 alright, this will speak to fewer people though

packages/mangrove.js/src/market.ts
```diff
@@ -247,6 +247,7 @@ class Market {
     base: string;
     quote: string;
     bookOptions?: Market.BookOptions;
+    noInit?: boolean;
```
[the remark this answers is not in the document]
adhusson: line 255
```typescript
    if (params["noInit"]) {
```

packages/mangrove.js/src/mangrove.ts
```diff
@@ -469,6 +487,186 @@ class Mangrove {
...
+    });
+  }
+
+  /**
+   * Returns open markets according to mangrove reader. Will internally update Mangrove token information.
+   *
+   * @param from: start at market i
+   * @param maxLen: max number of markets returned
+   * @param noInit: do not initialize markets (default:false)
+   * @param bookOptions: bookOptions argument to pass to every new market (default: undefined)
```
lnist, line 606:
```suggestion
   * @param from: start at market `from` (default: 0)
   * @param maxLen: max number of markets returned (default: all)
   * @param noInit: do not initialize markets (default: false)
   * @param bookOptions: bookOptions argument to pass to every new market (default: undefined)
```

lnist (changes requested):

lnist (approved):

PR #941, opened 2023-01-11
fix: correct package.json version (1.1.0)

adhusson:


lnist:
What happened to the release branch with this version change? :D

lnist (approved):

PR #953, opened 2023-01-23
fix!: do not poll logevents on every block

dontrolle:
Previously, semibooks listened for new blocks, then queried synchronously for events for that block, and then updated according to those events.

That processing was inefficient, and became very impractical over a websocket - where events only arrive one-by-one (but in order).

`semibook.ts` and `market.ts` have been rewritten to remove block-subscriptions and waits for the processing of events by block. Code using this have been rewritten to listen for log-events instead.

Test and test-utils have been rewritten to rely on logevents, as well.

Functionality and API in market and semibook supporting on-block listening has been removed as well as supporting code.


packages/mangrove.js/src/util/test/mgvIntegrationTestUtil.ts
```diff
@@ -250,9 +258,85 @@ export async function waitForBooksForLastTx(market?: Market) {
...
+    const asksCB = (
+      cbArg: Market.BookSubscriptionCbArgument,
+      bookEvent: Market.BookSubscriptionEvent
+    ) => {
+      if (cbArg.ba === "asks" && bookEvent.name === "SetGasbase") {
+        asksPromiseResolve();
+      }
+    };
+
+    let bidsPromiseResolve: (value?: any) => void;
```
adhusson, line 283: I think this could use `mangrove.js/src/util.ts` for clarity

packages/mangrove.js/src/liquidityProvider.ts
```diff
@@ -222,9 +222,29 @@ class LiquidityProvider {
...
+    const callback = (
+      cbArg: Market.BookSubscriptionCbArgument,
+      bookEvent: Market.BookSubscriptionEvent,
+      ethersLog: ethers.providers.Log
+    ) => {
+      const offerId = cbArg.offerId;
+      return { id: offerId, pivot: pivot, event: ethersLog };
+    };
+
+    const subscribeOnce = this.market
```
adhusson, line 236: I think this will trigger on the next OfferWrite received, whether it's the tx that was sent below (line 247) or not, because all it checks is `type == "OfferWrite"`? i.e. the caller of `newOffer` may get called back with a wrong offerId/pivot/event info, which makes the function unreliable. Whereas `onceWithTxPromise` takes the txPromise as an argument and makes sure it only triggers on that promise.

packages/mangrove.js/src/liquidityProvider.ts
```diff
@@ -222,9 +222,29 @@ class LiquidityProvider {
...
+      .once(
+        callback,
+        (cbArg: Market.BookSubscriptionCbArgument) => cbArg.type == "OfferWrite"
```
peterMangrove, line 239: Maybe this filter is not good enough. The "OfferWrite" event could come from another user. Just something to consider.

packages/mangrove.js/src/liquidityProvider.ts
dontrolle: Agree - I'd missed this. Follow-up commit d5a35fe38a596988766bde2aac00fc0e3a7dab46 should fix this, directly in `liquidityProvider.ts` 

Unless you protest, I still think removing `onceWithTxPromise()` from `market.ts` is a good idea. It had become quite complex, and was only used these few places in `liquidityProvider.ts` and in test-util code.

packages/mangrove.js/src/liquidityProvider.ts
dontrolle: Agree - I'd missed this. Follow-up commit d5a35fe38a596988766bde2aac00fc0e3a7dab46 should fix this, directly in `liquidityProvider.ts`

packages/mangrove.js/src/util/test/mgvIntegrationTestUtil.ts
dontrolle: I'm not sure I understand what part of `mangrove.js/src/util.ts` could be used here... Is it maybe `Deferred`, or?

I basically ended feeling more confident writing my own promise explicitly to state directly what I needed and didn't need.

adhusson: Yes sorry I meant to write Deferred but forgot! Fine either way :)

peterMangrove (approved):

packages/mangrove.js/src/liquidityProvider.ts
adhusson: Looks good re: checking the tx hash, but in this new version I'm not sure what happens if the tx reverts (or if the tx works but the log is never seen). Isn't the subscription checked forever in the market event processor?

packages/mangrove.js/src/liquidityProvider.ts
```diff
@@ -222,7 +222,39 @@ class LiquidityProvider {
...
+    });
+
+    const callback = async (
+      cbArg: Market.BookSubscriptionCbArgument,
+      bookEvent: Market.BookSubscriptionEvent,
+      ethersLog: ethers.providers.Log
+    ) => {
+      const txHash = (await txPromise).hash;
+      const logTxHash = ethersLog.transactionHash;
+      if (txHash === logTxHash && cbArg.type === "OfferWrite") {
```
peterMangrove, line 249: I guess multiple offers could be made in one transaction. Still not 100% sure that this will always get the correct offer.

peterMangrove (approved):

packages/mangrove.js/src/liquidityProvider.ts
adhusson: That problem also exists in the implementation this PR is replacing though. The expectation is that liquidity providers don't do crazy stuff on their `newOffer` function, but that's defiintely a weakness.

packages/mangrove.js/src/liquidityProvider.ts
dontrolle: Hmm, you're right. I wanted this out the door fast to unblock others, but this would introduce problems in those cases. I'll look at it with fresh eyes tomorrow and see if I can catch all corner-cases.

packages/mangrove.js/src/liquidityProvider.ts
dontrolle: Ok, I'll leave at it this.

packages/mangrove.js/src/liquidityProvider.ts
dontrolle: a4a5c62069ef7fb8d396287dd905d947d3fc357e handles the case where the tx fails, and adds a number of tests so as to prevent regression in this case.

packages/mangrove.js/test/integration/offermaker.integration.test.ts
```diff
@@ -262,6 +276,37 @@ describe("OfferMaker", () => {
         );
       });
 
+      it("fails, when trying to cancel a non-existing offer", async () => {
+        const retractPromise = onchain_lp.retractBid(666, true); // with deprovision
+
+        assert.rejects(
```
adhusson, line 282:
```suggestion
        await assert.rejects(
```

packages/mangrove.js/test/integration/offermaker.integration.test.ts
```diff
@@ -288,6 +333,31 @@ describe("OfferMaker", () => {
...
+        const { id: ofrId } = await onchain_lp.newBid({
+          wants: 10,
+          gives: 20,
+          fund: prov,
+        });
+
+        const base = onchain_lp.market.base.address;
+        const quote = onchain_lp.market.quote.address;
+        const closetx = await adminMgv.contract.deactivate(base, quote);
+        closetx.wait();
```
adhusson, line 349:
```suggestion
        await closetx.wait();
```

packages/mangrove.js/test/integration/offermaker.integration.test.ts
```diff
@@ -288,6 +333,31 @@ describe("OfferMaker", () => {
...
+
+        const updatePromise = onchain_lp.updateAsk(ofrId, {
+          wants: 12,
+          gives: 10,
+        });
+
+        assert.rejects(
```
adhusson, line 356:
```suggestion
        await assert.rejects(
```

packages/mangrove.js/test/integration/offermaker.integration.test.ts
```diff
@@ -262,6 +276,37 @@ describe("OfferMaker", () => {
...
+        const closetx = await adminMgv.contract.deactivate(base, quote);
+        closetx.wait();
+
+        const retractPromise = onchain_lp.retractBid(ofrId, true); // with deprovision
+
+        assert.rejects(
+          retractPromise,
+          "Retracting on a closed market should fail."
+        );
+      });
```
adhusson, line 308: Canceling on closed markets is supposed to work, but we can adapt it to "creating on a closed market fails"

```suggestion
      it("fails, when trying to create an offer on a closed market", async () => {

        const base = onchain_lp.market.base.address;
        const quote = onchain_lp.market.quote.address;
        const closetx = await adminMgv.contract.deactivate(base, quote);
        await closetx.wait();

        const prov = await onchain_lp.computeBidProvision();

        const createPromise = onchain_lp.newAsk({
          wants: 10,
          gives: 20,
          fund: prov,
        });

        await assert.rejects(
          createPromise,
          "Creating on a closed semibook should fail."
        );
      });
```

adhusson (commented):
added some awaits & adapted a test!

PR #1141, opened 2023-02-08
fix: better approve parameters

### LINKED ISSUE #562: `OfferLogic`'s `approveToken` and `approveMangrove` doesn't take overrides as expected
**Describe the bug**
`OfferLogic.approveToken("aDAI", overrides);` ignores the overrides.
Instead one should do:
`approveToken("aDAI", {}, overrides);` which is counter intuitive

This is due to ts not handling two optional arguments properly

┆Issue is synchronized with this [Monday item](https://mangrovedao.monday.com/boards/3894180513/pulses/3902026105) by [Unito](https://www.unito.io)

peterMangrove:
fixes #562

### DIFF
--- a/packages/mangrove.js/src/mgvtoken.ts
+++ b/packages/mangrove.js/src/mgvtoken.ts
@@ -15,6 +15,38 @@ namespace MgvToken {
   };
 }
 
+// Used to ease the user of approve functions
+export type ApproveArgs =
+  | Bigish
+  | ethers.Overrides
+  | { amount: Bigish; overrides: ethers.Overrides };
+
+function convertToApproveArgs(arg: ApproveArgs): {
+  amount?: Bigish;
+  overrides?: ethers.Overrides;
+} {
+  let amount: Bigish;
+  let overrides: ethers.Overrides;
+  if (arg["amount"]) {
+    amount = arg["amount"];
+  } else if (typeof arg != "object") {
+    amount = arg;
+  }
+  if (arg["overrides"]) {
+    overrides = arg["overrides"];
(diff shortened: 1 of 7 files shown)


packages/mangrove.js/src/mgvtoken.ts
```diff
@@ -15,6 +15,38 @@ namespace MgvToken {
   };
 }
 
+// Used to ease the user of approve functions
```
espendk, line 18:
```suggestion
// Used to ease the use of approve functions
```

packages/mangrove.js/src/mgvtoken.ts
```diff
@@ -146,25 +178,27 @@ class MgvToken {
...
-    return this.contract.approve(spender, _amount, overrides);
+    const args = convertToApproveArgs(arg);
+    const _amount =
+      "amount" in args
+        ? this.toUnits(args.amount)
+        : ethers.constants.MaxUint256;
+    return this.contract.approve(
+      spender,
+      _amount,
+      "overrides" in args ? args.overrides : {}
```
espendk, line 200: Perhaps this would be cleaner if `convertToApproveArgs` added `overrides: {}` when `overrides` isn't specified?

espendk (changes requested):
Nice! Added a few minor suggestions.

packages/mangrove.js/src/mgvtoken.ts
peterMangrove: Done

espendk (approved):

PR #1144, opened 2023-02-10
fix: use alchemy to get gas price

peterMangrove:


packages/bot-updategas/config/test.json
```diff
@@ -2,8 +2,7 @@
   "logLevel": "debug",
   "acceptableGasGapToOracle": 0.5,
   "constantOracleGasPrice": 6,
-  "runEveryXHours": 0.008,
-  "maxUpdateConstraint": {
-    "constant": 10
-  }
+  "network": "MATIC_MUMBAI",
+  "maxUpdateConstraint": {},
```
lnist, line 6: is the change of config of maxUpdateConstraint and runEveryXHours intentional? Then fine :)

packages/bot-updategas/config/default.json
```diff
@@ -1,9 +1,7 @@
 {
   "logLevel": "debug",
   "acceptableGasGapToOracle": 0.5,
-  "oracleURL": "https://gasstation-mainnet.matic.network/v2",
-  "oracleURL_Key": "standard",
-  "oracleURL_subKey": "maxFee",
+  "network": "MATIC_MUMBAI",
```
lnist, line 4: do we know how reliable alchemy's oracle is compared to the gas station? I am fine with the change if we think alchemy is at least as good. This change removes the option to use the old oracle - in theory we could have kept both options - but no reason to do that if alchemy is good enough.

lnist (approved):

packages/bot-updategas/config/test.json
peterMangrove: Copy paste mistake, reverted it.

packages/bot-updategas/config/default.json
peterMangrove: old oracle to specific. Alchemy has more oracles.

PR #1153, opened 2023-02-17
fix: approval

peterMangrove:


espendk:
Is this still relevant @peterMangrove or can we close it?

If still relevant, should target `develop` instead of `master`

peterMangrove:
@espendk lets just close it.

PR #1156, opened 2023-02-24
Adrien/deal for ob

adhusson:
the main purpose is that the custom orderbook can now deal with arbitrary tokens, on forks (and not, as before, only on tokens with permisionless minting).

ready for review 
has been decomposed into meaningful commits so review may be easier commit-by-commit


packages/mangrove.js/src/cli/commands/dealCmd.ts
```diff
@@ -0,0 +1,29 @@
...
+    {
+      spawn: false,
+      deploy: false,
+      host: argv.host,
+      port: argv.port,
+      pipe: true,
+    },
+    false
+  ).connect();
+  console.log("hey?");
```
peterMangrove, line 27: Should this be removed

packages/mangrove.js/src/script.ts
```diff
@@ -0,0 +1,15 @@
+import { Mangrove } from "./index";
+console.log("hellO?");
```
peterMangrove, line 2: Should this be removed?

peterMangrove (changes requested):

packages/mangrove.js/src/cli/commands/dealCmd.ts
adhusson: yes :D

packages/mangrove.js/src/script.ts
adhusson: as well!

adhusson: the whole file actually

PR #1170, opened 2023-03-15
chore/yargs dep

adhusson:
- mangrove.js v1.2.3-0 (#1158)
- feat(build): enable CI and PR validation on develop
- fix(build): Add empty string RPC urls (#1166)
- chore: add yargs as dependency


lnist (changes requested):
This looks like you are taking develop changes (including pre-release) to master, that seems wrong? Should the yargs dep change have been with a different base branch?

adhusson:
yes, thank you!

lnist (approved):

PR #1184, opened 2023-03-27
fix(test): Attempt fixing sporadic failing test by changing port number and killing process

lnist:


packages/mangrove.js/test/unit/mangrove.unit.test.ts
```diff
@@ -21,13 +21,11 @@ describe("Mangrove functionality", () => {
...
         ...defaultServerParams,
-        port: 8544,
+        port: 8544, // use port number below the one used in mochaHooks.ts
       }).connect();
 
-      // connect mgv
-      const provider = new ethers.providers.JsonRpcProvider(server.url);
       const mgv = await Mangrove.connect({
-        provider,
+        provider: server.url,
```
adhusson, line 28: I lack context as to what you are doing so I don't know what is going on here (replacing provider obj with url)

packages/mangrove.js/test/unit/mangrove.unit.test.ts
```diff
@@ -46,22 +44,23 @@ describe("Mangrove functionality", () => {
       ens["set(string,address)"]("Mangrove", ADDR1);
       await watcher.watchFor((k, v) => k === "Mangrove" && v == ADDR1);
       mgv.disconnect();
+      server.process.kill();
```
adhusson, line 47: Same here hard for me to have an opinion on that!

packages/mangrove.js/test/unit/mangrove.unit.test.ts
```diff
@@ -21,13 +21,11 @@ describe("Mangrove functionality", () => {
       // start server
       const server = await node({
         ...defaultServerParams,
-        port: 8544,
+        port: 8544, // use port number below the one used in mochaHooks.ts
```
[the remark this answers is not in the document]
lnist: No - the standard port is 8545 so this is below the one in mochahooks.

packages/mangrove.js/test/unit/mangrove.unit.test.ts
lnist: Instead of creating a provider here I simply use the normal code for it. Main difference is that mgv.disconnect will remove listeners.

packages/mangrove.js/test/unit/mangrove.unit.test.ts
lnist: I simply don't want the process to be hanging around - it is probably killed shortly due to the entire test run being complete. But for good measure I kill the anvil process as soon as we can.

PR #1199, opened 2023-04-04
Feat/semi book reorg resistant/louis

Louis-Amas:


packages/mangrove.js/src/tracker/blockManager.ts
```diff
@@ -0,0 +1,566 @@
...
+              subscriber.initializedAt
+            )} !== ${getStringBlock(cachedBlock)}`
+          );
+        }
+      }
+    }
+  }
+
+  /**
+   * Add new block in BlockManager cache detect reorganization and ensure that cache is consistent
```
espendk, line 474: I think some breaks/commas would be good here, eg:
```suggestion
   * Add new block in BlockManager cache, detect reorganization, and ensure that cache is consistent
```

packages/mangrove.js/src/util/sleep.ts
```diff
@@ -0,0 +1,4 @@
+export const sleep = (delayMs: number) =>
```
espendk, line 1: There's already a sleep function in https://github.com/mangrovedao/mangrove-ts/blob/develop/packages/commonlib.js/src/promiseUtil.ts

espendk (commented):
Submitting a partial review for now, will review more after the introduction Thursday :-)

dontrolle (approved):

PR #1228, opened 2023-04-27
fix(esbuild): fix resolving commonlib from multiple paths

lnist:
commonlib got resolved from multiple paths, and with the esbuild resolve plugin this only worked if the right import was hit first. That gave sporadic esbuild errors. This change uses absolute paths for the imports. Also removed testMaker shim as it was unused.

### DIFF
--- a/packages/mangrove.js/src/index.ts
+++ b/packages/mangrove.js/src/index.ts
@@ -23,7 +23,6 @@ import KandelFarm from "./kandel/kandelFarm";
 import KandelSeeder from "./kandel/kandelSeeder";
 import KandelInstance from "./kandel/kandelInstance";
 import OfferMaker from "./offerMaker";
-import TestMaker from "./util/test/TestMaker";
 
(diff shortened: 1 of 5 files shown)


packages/mangrove.js/src/index.ts
```diff
@@ -23,6 +23,7 @@ import KandelFarm from "./kandel/kandelFarm";
 import KandelSeeder from "./kandel/kandelSeeder";
 import KandelInstance from "./kandel/kandelInstance";
 import OfferMaker from "./offerMaker";
+import TestMaker from "./util/test/TestMaker";
```
espendk, line 26:
```suggestion
```

espendk (changes requested):
lgtm.

As just discussed IRL, `import.meta.url` could be problematic in browser environments (at least we've seen similar issues in the past). But since this script is only for the build, this is not a concern here.

espendk (approved):

PR #1229, opened 2023-04-27
feat(test): Run tests in parallel

lnist:
* Each anvil runs on its own port, separated by 1000
* Proxies add to the port number (will break at 1000 tests)
* isRejected did not work (` TypeError: chai_1.assert.isRejected is not a function` for some reason)
* Retry on deploy failures due to https://github.com/foundry-rs/foundry/issues/2884
* Wait for pending txs in anvil to hopefully fix sporadic failing tests
* Close proxies at end to avoid issues on parallel runs
* Possibly fix a race condition in a test
* Note: Not used for CI (test:coverage) since action runners are so limited it doesn't change anything.

`time yarn test:integration` goes from ~27s to ~13-17s


packages/mangrove.js/src/util/test/mochaHooks.ts
```diff
@@ -67,20 +97,45 @@ export const mochaHooks = {
...
+        throw new Error("Missing txpool data");
+      }
+
+      if (
+        Object.keys(result.pending).length ||
+        Object.keys(result.queued).length
+      ) {
+        console.log("txpool_content not empty... waiting...");
+        console.log(JSON.stringify(result));
+        await sleep(200);
```
adhusson, line 131: why is this needed?

packages/mangrove.js/test/integration/kandel.integration.test.ts
```diff
@@ -1445,8 +1445,8 @@ describe("Kandel integration tests suite", function () {
           );
 
           // Assert
-          assert.isUndefined(approvalTxs2[0]);
-          assert.isUndefined(approvalTxs2[1]);
+          assert.equal(approvalTxs2[0], undefined);
```
adhusson, line 1448: are those tests changes related to the "run tests in parallel" ?

packages/mangrove.js/test/integration/mangroveToyENS.integration.test.ts
```diff
@@ -41,8 +56,11 @@ describe("Mangrove functionality", () => {
 
       // watch for new entry
       const ADDR1 = "0x0000000000000000000000000000000000000001";
+      const watchPromise = watcher.watchFor(
+        (k, v) => k === "Mangrove" && v == ADDR1
+      );
       ens["set(string,address)"]("Mangrove", ADDR1);
-      await watcher.watchFor((k, v) => k === "Mangrove" && v == ADDR1);
+      await watchPromise;
```
adhusson, line 63: Confused by this change

adhusson:
Really cool PR 😍

packages/mangrove.js/src/util/test/mochaHooks.ts
lnist: Due to https://github.com/foundry-rs/foundry/issues/4843 - I am not sure what happens if we call evm_revert with pending transactions. Adding this check seems to fix sporadic failing tests .

packages/mangrove.js/test/integration/kandel.integration.test.ts
lnist: Yes! I had to remove chai import since isRejected does not work for some reason. Changed to use assert instead, and that did not have .isUndefined.

packages/mangrove.js/test/integration/mangroveToyENS.integration.test.ts
lnist: I start watching before you call ens. Otherwise ens may have finished its stuff before we set up watching, so we don't get the change.

adhusson (approved):

PR #1238, opened 2023-05-02
Feat/completes offer logics

jkrivine:
This PR does the following:
* completes the `offerLogic` class for all public functions of `IOfferLogic`. In particular the semantics of `logic.approve` and `logic.allowance` is now clearer: the owner is the contract.
* removes `approveAsks` and `approveBids` from the `LiquidityProvider` class. Those functions were making the wrong assumption that the existence of a router implies that makers must approve the router for transfer.
* adds abstraction for `transfer` and `transferFrom` to the `mgvToken` class


peterMangrove (changes requested):
lgtm, just the one import should be removed. 
And possibly some tutorials that needs to be updated because of this.

packages/mangrove.js/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# next
+ 
+# 1.2.4-11 (may 2023)
```
peterMangrove, line 3: Not sure you need to bump the offer version, i should just be #next. @lnist said something like this.

peterMangrove (changes requested):

PR #1263, opened 2023-05-12
Feat/fix approve

jkrivine:
This PR does the following (and replaces #1238 )

- OfferLogic has an `approve` function to ask the underlying logic to approve signer (or an arbitrary address) to spend a specific token on its behalf.
- Liquidity provider no longer provides `approveAsk`, `approveBids` which were making too much asumptions on router usage.
- Adapts tutorial scripts accordingly.
- bugfix: token approval could not be set to 0

### DIFF
--- a/packages/mangrove.js/test/integration/offermaker.integration.test.ts
+++ b/packages/mangrove.js/test/integration/offermaker.integration.test.ts
@@ -84,8 +84,9 @@ describe("OfferMaker", () => {
 
     describe("Before setup", () => {
       it("checks allowance for onchain logic", async () => {
+        const logic = onchain_lp.logic as OfferLogic;
         let allowanceForLogic /*:Big*/ = await mgv.token("TokenB").allowance({
-          owner: onchain_lp.logic.address,
+          owner: logic.address,
           spender: mgv.address,
         });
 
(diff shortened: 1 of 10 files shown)


jkrivine:
> lgtm. When this hits the docs repo, then some guides should be updated. Tagging @dontrolle just as an FYI

@dontrolle I updated the scripts used for the tutorials, I have not found anything in the docs that needs to be updated

jkrivine:
@peterMangrove I've added a regression test that fails. I don't understand why yet.

packages/mangrove.js/test/integration/offermaker.integration.test.ts
```diff
@@ -356,6 +360,19 @@ describe("OfferMaker", () => {
...
+        const base = onchain_lp.market.base;
+        const logic = onchain_lp.logic as OfferLogic;
+
+        const tx = await logic.approve(base.name, { optAmount: 42 });
+        const resp = await tx.wait();
+        let allowance = await base.allowance({
+          owner: logic.address,
+          spender: await logic.mgv.signer.getAddress(),
+        });
+        assert.equal(allowance, 42, "Invalid allowance");
```
peterMangrove, line 374: The allowance is of type "Big", i don't think big and number can be compared.

peterMangrove (changes requested):

packages/mangrove.js/test/integration/offermaker.integration.test.ts
jkrivine: Seems to work at leat :)

peterMangrove (approved):

PR #1267, opened 2023-05-15
feat(eslint): Enable warnings for missing/extra async/await for mangrove.js

lnist:


packages/mangrove.js/test/integration/kandel.integration.test.ts
```diff
@@ -1589,7 +1597,10 @@ describe("Kandel integration tests suite", function () {
           await waitForTransactions(kandel.populate({ distribution }));
         });
 
-        [{ factor: 0.5 }, { gasreq: 1 }].forEach(({ factor, gasreq }) => {
+        [
+          { factor: 0.5, gasreq: undefined },
+          { factor: undefined, gasreq: -10000 },
```
peterMangrove, line 1602: Is this on purpose? seems gasreq was not set before?

peterMangrove (commented):
lgtm 🔥 Only that one small comment. When this hits develop, you should probably make a statement on slack, saying that you have enabled this. Doubt anyone will disagree.

Going to be very nice to have this enabled!

packages/mangrove.js/test/integration/kandel.integration.test.ts
lnist: Either factor was undefined, or gasreq was undefined before (two separate objects in the array) - it is just explicit now since I was confused :) . Also, it should throw an exception on too low density, and I only got that by using a very low gasreq due to very low density requirements in the default setup.

PR #1291, opened 2023-05-23
Fix/UI semibook not updating

Louis-Amas:


espendk (approved):
👍

.github/workflows/publish-commonlib-js.yml
```diff
@@ -100,7 +100,7 @@ jobs:
         export RELEASE_NOTES=`sed -n "/^# $PACKAGE_VERSION/,/^#/p" CHANGELOG.md | grep -v "^#" | sed -e :a -e '/./,$!d;/^\n*$/{$d;N;};/\n$/ba'`
         echo $RELEASE_NOTES
         gh release create "v$PACKAGE_VERSION" --title "v$PACKAGE_VERSION" --notes "$RELEASE_NOTES"
-        gh pr create --title "v$PACKAGE_VERSION" --body "$RELEASE_NOTES" --label "release" --reviewer "mangrovedao/all-devs"
+        gh pr create --title "v$PACKAGE_VERSION" --body "$RELEASE_NOTES" --reviewer "mangrovedao/all-devs"
```
adhusson, line 103: Should this be part of the commit?

.github/workflows/publish-mangrove-js.yml
```diff
@@ -59,13 +59,9 @@ jobs:
...
         fi
-        echo "npmAuthToken: \"$NPM_TOKEN\"" >> ~/.yarnrc.yml
-        cd packages/mangrove.js
 
-        echo 'yarn'
-        yarn --version
-        cat ~/.yarnrc.yml
-        cat ~/.yarnrc.yml >> .yarnrc.yml
+        echo "npmAuthToken: \"$NPM_TOKEN\"" >> .yarnrc.yml
+        cd packages/mangrove.js
```
adhusson, line 64: Same as above, not sure it should be part of the commit

packages/mangrove.js/src/market.ts
```diff
@@ -730,13 +749,21 @@ class Market {
...
+        cb,
+        params,
+      });
     });
   }
 
   /* Stop calling a user-provided function on book-related events. */
   unsubscribe(cb: Market.StorableMarketCallback): void {
-    this.#subscriptions.delete(cb);
+    const index = this.#subscriptions.findIndex((el) => el.cb === cb);
```
adhusson, line 761: The complexity has changed here since we're moving from a ~constant map access to iterating over the callbacks array

PR #1305, opened 2023-05-25
feat(Kandel): Allow price points to be generated outwards from midPrice with a hole

lnist:
Previously prices where generated from minPrice and upwards by multiplying by ratio until we hit maxPrice.
Now, if a midPrice is supplied, we generated outwards from that until we hit minPrice and maxPrice, with a hole at the midPrice.

Note: this required additional changes since multiple places did not support a hole in the list of prices, and for instance the initial provision for a distribution should be for the entire list, and not disregard the hole.

### DIFF
--- a/packages/mangrove.js/src/kandel/kandelPriceCalculation.ts
+++ b/packages/mangrove.js/src/kandel/kandelPriceCalculation.ts
@@ -1,17 +1,19 @@
 import Big from "big.js";
 import { Bigish } from "../types";
 
-/** Parameters for calculating a geometric price distribution. Exactly three must be provided.
+/** Parameters for calculating a geometric price distribution. Exactly three of minPrice, maxPrice, ratio, and pricePoints must be provided.
  * @param minPrice The minimum price in the distribution.
  * @param maxPrice The maximum price in the distribution.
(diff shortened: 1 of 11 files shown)


packages/mangrove.js/src/kandel/kandelPriceCalculation.ts
```diff
@@ -121,18 +127,37 @@ class KandelPriceCalculation {
...
+        );
+      }
+    };
+
     let price = minPrice;
+    if (midPrice) {
+      price = midPrice.div(ratio);
+      while (price.gte(minPrice)) {
+        prices.push(price);
+        checkPricesLength();
```
peterMangrove, line 146: I guess you calculate this the number of price points, with a formula and then only do this check once. Not super important though

peterMangrove (approved):
Lgtm, only one small suggestion

packages/mangrove.js/src/kandel/kandelPriceCalculation.ts
lnist: Only approximately, so we need the hard check, so I'll leave it.

PR #1309, opened 2023-05-26
fixing NotEnoughGasForMaker reverts

jkrivine:
This PR increases gasEstimates of market.ts by 10%. This is to allow a buffer of underestimation is gasbase which is not precise.

### DIFF
--- a/packages/mangrove.js/CHANGELOG.md
+++ b/packages/mangrove.js/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Next version
 
 - Use array to handle market subscription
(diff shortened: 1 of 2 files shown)


packages/mangrove.js/test/integration/market.integration.test.ts
```diff
@@ -1479,6 +1479,7 @@ describe("Market integration tests suite", () => {
       "buy",
       BigNumber.from(1)
     );
+    console.log(emptyBookAsksEstimate.toString());
```
espendk, line 1482: Should this be here?

jkrivine: nope

espendk (approved):
lgtm

PR #1310, opened 2023-05-26
feat: Remove commonlib.js

dontrolle:
bot-utils now also depends on a specific version of the mangrove.js package


espendk (approved):
👍

dontrolle:
This build https://github.com/mangrovedao/mangrove-ts/actions/runs/5090466993 ensures that this is actually a green build. (There are workflow changes.)

PR #1319, opened 2023-05-30
fix(openMarkets): Use mgv name for tokens instead of symbol

lnist:


adhusson (approved):
lgtm :)

lnist:
> Is the `name` here related to the ERC20 `name`? https://ethereum.org/en/developers/docs/standards/tokens/erc-20/#methods

No, it is related to `token(name: string, ...)` in mangrove.ts which is our internal name which is the key in addresses and context. It is most often the same as symbol, but may not be.

espendk:
ok. I think it's a bit unfortunate to use `name` for something else than
the standard name in a class that represents an erc20

tir. 30. maj 2023 11.41 skrev Lasse ***@***.***>:

PR #1329, opened 2023-06-01
feat(gas): Extend simulateMarketOrder to use it for gas estimation

lnist:
The change is two commits:
1) Use real density and gasbase in tests to prepare for the next commit
2) Extend simulateMarketOrder to use it for gas estimation

Notes:
* Since the estimation is for a market order I currently add the imperical (for tests) 200.000 extra gas when MangroveOrder is used.
* Since offers can fail during real execution, and market can change, I opted to add 50% extra.

This replaces the estimateGas function which can be deleted. The estimate gas assumes that density is a good proxy for gas requirement, and it uses the getMaxGasReq function which reads through the entire semibook (not only cache). Since we have built the simulation function, it seemed fitting to use that for better estimation. Also, we previously did not account for MangroveOrder taking extra gas.


adhusson:
A general comment that comes to mind is that `estimateGas` abstracts the actual offers away to be robust against asynchrony issues. Here the simulation (even with a % overhead added) will depend on the current book state, which we hope will change many times during a block (as in be consumed & refilled). I'm wondering if there is a third way that gives a safer gas estimate but is better than the current `estimateGas`.

adhusson (approved):

PR #1330, opened 2023-06-01
fix: restore deal cmd, prevent future mistyping, add required options

adhusson:


packages/mangrove.js/src/cli/mgv.ts
```diff
@@ -9,12 +9,14 @@ import * as dealCmd from "./commands/dealCmd";
...
 
+type StrictCM = yargs.CommandModule & { builder: (...args: any[]) => any };
+
 yargs
-  .command(parrotCmd)
-  .command(printCmd)
-  .command(retractCmd)
-  .command(dealCmd) // note: node subcommand env vars are prefixed with MGV_NODE instead of MGV_
-  .command(nodeCmd) // note: node subcommand env vars are prefixed with MGV_NODE instead of MGV_
+  .command(parrotCmd as StrictCM)
```
espendk, line 15: Will this give a compile error if `parrotCmd` doesn't match the `StrictCM` type? In that case, nice :-)

espendk (commented):
good catch!
I couldn't find documentation that `a as B` will throw if `a` does not match `B` is that the case? Or how do the `as StrictCM` expressions prevent the problem in the future?

packages/mangrove.js/src/cli/mgv.ts
adhusson: After trying again, sadly it only catches situations where no `builder` field is exported. It doesn't catch a builder of the wrong type. The error `error TS2352: Conversion of type 'typeof import(...)' to type 'StrictCM' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.` made me think it would catch wrong types but no.

espendk (approved):

PR #1345, opened 2023-06-08
feat(gas): Make gas estimation explicit in orders to allow usage of ethers estimate

lnist:
* fix(offerLogic): Addresses should be checksum addresses when used in comparisons
* feat(trade): Extract getRawParams
* feat(gas): Make gas estimation explicit in orders to allow usage of ethers estimate
* fix(snipe): Do not use gas estimate when sniping
* feat(market): Expose gas estimation functions

### DIFF
--- a/src/util/trade.ts
+++ b/src/util/trade.ts
@@ -118,6 +118,35 @@ class Trade {
     return this.comparePrices(price, priceComparison, referencePrice);
   }
 
+  getRawParams(bs: Market.BS, params: Market.TradeParams, market: Market) {
+    const { gives, wants, fillWants } =
+      bs === "buy"
+        ? this.getParamsForBuy(params, market.base, market.quote)
+        : this.getParamsForSell(params, market.base, market.quote);
+    const restingOrderParams =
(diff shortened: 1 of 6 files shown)


src/util/trade.ts
```diff
@@ -273,6 +304,59 @@ class Trade {
...
+      params,
+      market
+    );
+    const ba = this.bsToBa(bs);
+
+    switch (orderType) {
+      case "restingOrder":
+        // add an overhead of the MangroveOrder contract on top of the estimated market order.
+        return (await market.simulateGas(ba, gives, wants, fillWants)).add(
+          200000
```
espendk, line 333: This constant is used at least twice, maybe extract it?

espendk (commented):
great! Will be interesting to see the effect on the web app

Small comment re. the MangroveOrder overhead constant that is unnamed and used in two places: Would probably be good to make it a named constant.

src/util/trade.ts
lnist: done

espendk (approved):

PR #1367, opened 2023-06-14
chore(ts): Enable strict

lnist:


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "allowSyntheticDefaultImports": true,
     "emitDecoratorMetadata": false, // this is the default. Calling it out since esbuild does not support it.
     "target": "es2020",
-    //"strict": true,
+    "strict": true,
     "strictNullChecks": true,
```
espendk, line 11: Is `strictNullChecks` included in `strict`?

lnist: Correct. I've removed that one and noImplicitAny

PR #1372, opened 2023-06-15
feat: Make mangrove.js easily configurable

espendk:
Configuration can now be extended and/or changed via the `Mangrove.updateConfiguration(partialConfiguration)` method. The provided `partialConfiguration` will be merged into the existing configuration, overriding any configuration that occurs in both. See the `Configuration` type for the structure of the configuration.

Example, adding configuration for a new token with symbol "SYM":
```TS
  Mangrove.updateConfiguration({
    tokens: {
      SYM: {
        decimals: 18
      }
    }
  })
```

A few related improvements are included:
- Renamed `constants.ts` to `configuration.ts` and encapsulated all configuration there

- Token decimals handling and API have been improved:
  - `Mangrove.token(.)` is now async and will fetch decimals from chain if they are not in the configuration. This makes it more robust with respect to tokens that are not included in the configuration of mangrove.js. The old sync behaviour that relied only on the configuration (and failed if the decimals were unknown) is still available in `Mangrove.tokenFromConfig(.)`.
  - `MgvToken.getDecimals(.)` etc. will now return `undefined` if the decimals are not in the configuration instead of throwing an `Error`. The old behaviour is still available in the new `MgvToken.getDecimalsOrFail(.)`.
  - `MgvToken.getOrFetchDecimals(.)` will first look for decimals in the configuration and then, if they are not found, fetch them from chain.

- `KandelConfiguration`:
  - There are now types that describe the structure of Kandel configuration
  - Overrides can now be partial as they are applied on top of the configuration in `configuration` instead of replacing it.


adhusson (approved):
lgtm, but e.g. @lnist should look at the Kandel part. Also could use some minimal doc (e.g. one example) either in the README or in mangrove.ts!

espendk:
> Also could use some minimal doc (e.g. one example) either in the README or in mangrove.ts!

I've slightly expandeded and TSDoc'ified the documentation of `Mangrove.updateConfiguration`.

The README doesn't have any API documentation, so there's no natural place there to add it. Instead, we should add it to the dev documentation in mangrove-docs when this is PR is merged and published.

(Reviewing the README, I found some deprecated monorepo documentation that I've removed)

src/configuration.ts
```diff
@@ -13,334 +15,485 @@ import {
...
+/** Kandel configuration for a specific market.
+ * @param aaveEnabled Whether AaveKandel should be allowed to be used.
+ * @param minimumBasePerOfferFactor Additional factor for the minimum amount of base token that should be offered per offer to stay above density requirements.
+ * @param minimumQuotePerOfferFactor Additional factor for the minimum amount of quote token that should be offered per offer to stay above density requirements.
+ * @param spread The default spread used when transporting funds from an offer to its dual.
+ * @param ratio The default ratio of the geometric progression of prices.
+ */
+export type KandelMarketConfiguration = {
+  aaveEnabled: boolean;
+  minimumBasePerOfferFactor: Big;
```
lnist, line 89: Are these really Big? I am not sure how it is converted into the type. I would expect these to be string, or Bigish?

PR #1436, opened 2023-07-13
Fix/limit order pivot

jkrivine:
This PR fixes a bug that led to a wrong pivot computation when running limit orders.
It also passes `pivotId` to restingOrder results in order to be able to perform regression tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Next version
 
-# 1.4.10
-
 - fix: wrong pivot when posting limit orders far from the mid price.
 
 # 1.4.9
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next version
 
+# 1.4.10
+
+- fix: wrong pivot when posting limit orders far from the mid price.
+
 # 1.4.9
 
 - Bump: reliable-event-subscriber to v1.1.24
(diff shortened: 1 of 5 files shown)


src/util/trade.ts
```diff
@@ -566,8 +566,8 @@ class Trade {
     const { outbound_tkn, inbound_tkn } = market.getOutboundInbound(ba);
     const price = Market.getPrice(
       ba,
-      inbound_tkn.fromUnits(gives),
-      outbound_tkn.fromUnits(wants)
+      /* (maker) gives is takerWants*/ outbound_tkn.fromUnits(wants),
+      /* (maker) wants is takerWants*/ inbound_tkn.fromUnits(gives)
```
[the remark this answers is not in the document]
peterMangrove: Okay, but it says "takerWants" in both line. That seems wrong

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next version
 
+# 1.4.10
```
peterMangrove, line 3: The release workflow will add this. Remove this line

peterMangrove (changes requested):

CHANGELOG.md
jkrivine: even if I publish with the UI ?

peterMangrove: I'm not entirely sure what you mean by UI. If you mean running the workflow using the UI, then yes. I'm 100% sure, but i think it will bump it to 1.14.11 in the changelog. We can try, and see what happens

peterMangrove (approved):

PR #1531, opened 2023-10-04
fix: infinite approval checks for 2^200

peterMangrove:
This PR changes the infinite approvals check from checking against 2^256, to instead check if the allowance is larger than 2^200. This allows for a broader span of "infinite allowance", this is to handle allowance slowly being used when user do Limit and/or Market orders.


johanleroch (approved):
Thanks for this change!

PR #1568, opened 2023-10-27
fix(tests): mochaHooks

Louis-Amas:


dontrolle (approved):

src/util/forgeScript.ts
```diff
@@ -13,7 +13,6 @@ export function execForgeCmd(
   // Foundry needs these RPC urls specified in foundry.toml to be available, else it complains
   env = {
```
lnist, line 14: Is there a need to copy env to itself? Now comment also seems even more stale.

foundry.toml
```diff
@@ -23,12 +23,6 @@ testnet_zkevm="https://rpc.public.zkevm-test.net"
 zkevm="https://zkevm-rpc.com"
 goerli="${GOERLI_NODE_URL}"
 
-[etherscan]
-goerli={key="${GOERLI_API_KEY}"}
-mumbai={key="${POLYGON_API_KEY}"}
```
lnist, line 28: Have you verified locally that it works if you do not have these defined in a .env file or your environment? (I expect it to work, just want to make sure)

src/util/test/mochaHooks.ts
```diff
@@ -120,7 +120,9 @@ export const mochaHooks = {
   },
 
   async beforeEachImpl(hook: hookInfo) {
-    Mangrove.setAddress("OfferMaker", hook.offerMakerAddress!, "local"); //FIXME: have to set the address in beforeEach, instead of beforeAll, because mangrove configuration gets reset by other tests. (e.g. configuration unit tests)
+    if (hook.offerMakerAddress) {
```
lnist, line 123: This is still a wrong fix. We should either make sure offerMakerAddress is always set (I thought you fixed that in bots?) or we should move the offerMaker code to the test where it belongs.

lnist (changes requested):

lnist (approved):

PR #1572, opened 2023-10-30
fix(liquidityProvider): tick caclulation

Louis-Amas:


src/liquidityProvider.ts
```diff
@@ -239,10 +239,7 @@ class LiquidityProvider {
     } else {
       gives = Big(p.gives);
       let [base_amt, quote_amt] = [gives, Big(p.wants)];
-      if (p.ba === "bids") {
-        [base_amt, quote_amt] = [quote_amt, base_amt];
-      }
-      price = Big(quote_amt).div(base_amt);
+
       tick = TickLib.tickFromVolumes(
```
lnist, line 243: Hm, this looks wrong now actually. tickFromVolumes works on raw BigInteger values - that is, it does not consider decimals, so line 244 converts based on base and quote decimals.

However, gives will be in base for an ask and quote for a bid, so the correct decimals should be used based on that. I would not use the "baseAmount and quoteAmount" (please rename though I know you did not introduce those names) variables.

test/unit/liquidityProvider.unit.test.ts
```diff
@@ -111,7 +111,7 @@ describe("Liquidity provider unit tests suite", () => {
     assert.equal(price.toFixed(4), "0.6667");
     assert.deepStrictEqual(gives, Big(20));
     assert.deepStrictEqual(fund, undefined);
-    assert.deepStrictEqual(tick.toNumber(), 280378);
```
lnist, line 114: IF the price was basically inversed before, then I woul dhave expected it to change from 280378 to -280378. Maybe the reason is the decimals issue?

lnist (changes requested):

src/liquidityProvider.ts
```diff
@@ -238,19 +238,26 @@ class LiquidityProvider {
...
         BigNumber.from(
-          base_amt.mul(Big(10).pow(market.base.decimals)).toFixed()
+          p.ba === "asks"
+            ? wants.mul(Big(10).pow(market.quote.decimals)).toFixed()
+            : wants.mul(Big(10).pow(market.base.decimals)).toFixed()
         ),
         BigNumber.from(
-          quote_amt.mul(Big(10).pow(market.quote.decimals)).toFixed()
+          p.ba === "bids"
```
lnist, line 250: I would switch this if to be using asks just like the other if... and it makes me wonder why tests are green, because as it stands now it seems base and quote decimals are wrong here.

lnist (changes requested):

lnist (approved):

PR #1574, opened 2023-11-06
feat(upgrade): Upgrade to newest mangrove-strats

lnist:
Main change is that offerGasreq is removed.

For now, we define the configuration per network, but we will in the future need pass in arbitrary gasreq for arbitrary underlying routers. I suggest we postpone that change till we have the need.

The chosen configuration is as it was defined in solidity prior to this change (except I add 1 for the mumbai values to use in tests)


espendk (approved):
lgtm

PR #1578, opened 2023-11-07
feat(TickPriceHelper): set market and market-side in ctor

dontrolle:


src/util/tradeEventManagement.ts
```diff
@@ -51,14 +49,15 @@ class TradeEventManagement {
     const id = this.#rawIdToId(raw.id);
 
     if (id === undefined) throw new Error("Offer ID is 0");
+    const tickPriceHelper: TickPriceHelper = new TickPriceHelper(ba, market);
```
lnist, line 52: seems a bit overkill to instantiate for each event. Consider putting TickPriceHelper-instance on semibook?

dontrolle: Good point, and better abstraction. Revising.

dontrolle: Updated in e252c680127819e5222eac6f9f0b868cf6eb55ae

lnist (dismissed):

lnist (approved):

PR #1579, opened 2023-11-08
feat: Use addresses from mangrove-deployments and context-addresses packages

espendk:
This PR is a first improvement to the deployment process: It moves context addresses out of this repo and instead reads addresses from these new repos/npm packages:

- Deployments: [mangrovedao/mangrove-deployments](https://github.com/mangrovedao/mangrove-deployments/)
- Context addresses: [mangrovedao/context-addresses](https://github.com/mangrovedao/context-addresses/).

See the companion PR's in `mangrove-core` and `mangrove-strats` for more details:

- https://github.com/mangrovedao/mangrove-core/pull/651
- https://github.com/mangrovedao/mangrove-strats/pull/120

## Brief solution overview
The configuration loading logic has been changed to load addresses from the above npm packages instead of from `mangrove-core`, `mangrove-strats`, and the `addresses.json` file.

When querying `mangrove-deployments` for deployment addresses, one must specify the contract version and whether the deployment is of a "released" version.

`configuration.ts` queries for contract versions matching patches of the package versions listed in `package.json` for `mangrove-core` and `mangrove-strats`, regardless of whether they are released. Later, we may wish to provide a way for clients to control this behaviour.


src/configuration.ts
```diff
@@ -500,44 +500,105 @@ export function resetConfiguration(): void {
...
 
-  for (const [network, networkAddresses] of Object.entries(
-    loadedAddressesByNetwork
-  )) {
-    for (const [name, address] of Object.entries(networkAddresses) as any) {
-      if (address) {
-        addressesConfiguration.setAddress(name, address, network);
-      }
+function readMangroveDeploymentAddresses() {
+  // FIXME: Consider how to expose other deployments than the primary
```
dontrolle, line 511: I'd lean towards not committing these 3 FIXME's. When I see a FIXME, I think "a known bug or problem". 

What is committed here works, and is consistent. We could add more configurability or degrees of freedom, but I'd say that is future features, and not FIXME-worthy. At the very least, I'd downgrade these to a simple comment-note.

dontrolle (changes requested):

dontrolle (approved):
🌴

PR #1589, opened 2023-11-13
Reintroduce market offer volume

dontrolle:


test/integration/market.integration.test.ts
```diff
@@ -550,6 +556,7 @@ describe("Market integration tests suite", () => {
       const mockedMarket = mockito.spy(market);
       const semiBook = mockito.mock(Semibook);
       const ba = "asks";
+      const expectedGives = new Big(-12);
```
lnist, line 559: testing with a negative value seems weird... gives is a uint

dontrolle: Hmmm, I tend to agree. I didn't introduce those values (simply refactored the test), but I'll change it to something reasonable.

dontrolle: Fixed in 46f17c50cda49ccd1d3f8b38f5ef9f7a4bbedad9

espendk (approved):
nice 👌

PR #1596, opened 2023-11-16
feat(semibook) Reintroduce optional boundary on VolumeParams

dontrolle:
NOTE: Non-functional, right now.

### DIFF
--- a/src/semibook.ts
+++ b/src/semibook.ts
@@ -54,6 +54,9 @@ namespace Semibook {
     given: Bigish;
     /** Whether `given` is base to be bought or quote to be sold. */
     to: Market.BS;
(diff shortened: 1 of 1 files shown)


src/semibook.ts
```diff
@@ -54,6 +54,9 @@ namespace Semibook {
     given: Bigish;
     /** Whether `given` is base to be bought or quote to be sold. */
     to: Market.BS;
+    /** Optional: induce a max avg. price after which to stop buying/selling.
```
lnist, line 57:
```suggestion
    /** Optional: induce a max price after which to stop buying/selling.
```

lnist (approved):

PR #1597, opened 2023-11-16
fix(test): Use gasprice 1000Mwei via upgrade of strats package and various minor fixes

lnist:
* Depends on https://github.com/mangrovedao/mangrove-strats/pull/136 and a package release.
* Uses gasprice 1000MWei instead of 1MWei (used to be 1gwei) - causes updates to some tests since using gasprice lower than global, e.g., for creating offers will use mangrove's global instead.
* Downgrades prettier until we can fix config or do bulk update of files.
* Adds some OLKey helpers - can be used more widespread in later PRs


dontrolle (approved):
lgtm

dontrolle (approved):

PR #1606, opened 2023-11-17
feat: Introduce and use TickPriceHelper.TickFromVolumes

dontrolle:


src/liquidityProvider.ts
```diff
@@ -219,36 +217,25 @@ class LiquidityProvider {
   } {
     const tickPriceHelper = new TickPriceHelper(p.ba, market);
     let tick: ethers.BigNumber, gives: Big, price: Big;
-    // deduce price from tick & gives, or deduce tick & gives from volume & price
     if ("tick" in p) {
+      // deduce price from tick & gives
       tick = ethers.BigNumber.from(p.tick);
       price = tickPriceHelper.priceFromTick(tick);
```
lnist, line 223: I would delete some code here... normalizeofferparams is never used to get the price, so remove the price calculation completely.

PR #1613, opened 2023-11-21
feat: create tick price helper for each market

Louis-Amas:


lnist (commented):
You should already be able to do market.getBook().asks.tickPriceHelper if you have a market.

PR #1614, opened 2023-11-23
fix(semibook): use tickPriceHelper

Louis-Amas:


Louis-Amas:
Closing because similar works is done here https://github.com/mangrovedao/mangrove.js/pull/1615

PR #1617, opened 2023-11-23
feat(kandel): Split Geometric Kandel from core kandel and allow usage of specialized populate function

lnist:
Most is just refactoring - splitting out geometric Kandel-specific parts. The real changes is the addition of the populateGeometricDistribution in geometricKandelInstance.ts

Also note the changes to test configuration.


espendk (approved):
lgtm

PR #1620, opened 2023-11-24
Fix/louis/semi book tick

Louis-Amas:


package.json
```diff
@@ -97,7 +97,7 @@
     "@types/node": "^20.3.1",
     "@types/node-cleanup": "^2.1.2",
     "@types/shelljs": "^0.8.11",
-    "@types/yargs": "^17.0.12",
+    "@types/yargs": "^17.0.32",
```
lnist, line 100: why upgrade yargs?

src/semibook.ts
```diff
@@ -1381,7 +1389,11 @@ export class SemibookCacheOperations {
   // remove offer id from book and connect its prev/next.
   // return 'undefined' if offer was not found in book
   // This modifies the cache so must be called in a context where #cacheLock is acquired
-  removeOffer(state: Semibook.State, id: number): Market.Offer | undefined {
+  removeOffer(
```
lnist, line 1392: this change seems unrelated - do we have tests that exercise this situation? (I am not too familiar with the semibook cache)

test/integration/semibook.integration.test.ts
```diff
@@ -1231,8 +1262,11 @@ describe("Semibook integration tests suite", function () {
               chunkSize: 1, // Fetch only 1 offer in each chunk
             },
           });
-          const semibook = market.getSemibook("asks");
-          expect(semibook.size()).to.equal(4); // need 4 offers, as each offer gives slightly less than one, do to tick.
+          const semibook2 = market2.getSemibook("asks");
+
+          market2.consoleAsks();
```
lnist, line 1267: remove

lnist (commented):
I have not reviewed all the test changes in detail, they seem to follow an overall pattern, so as long as you didn't suddenly get completely different results, then all good there.

package.json
Louis-Amas: Good question, it happened without my consent, should I remove it ?

src/semibook.ts
Louis-Amas: It's related as it permit to make this test pass "returns correct estimate and residue when cache is partial and insufficient while offer list is sufficient" :)

When the cache reached is max capacity and that we fetch more offers to fulfill the `estimateVolume` call, we first insert the offer newly fetched offer and then we remove it from the cache (as the cache is full). But when we remove it we were previously removing the next pointer to this offer. The cache should stay aware that there exists offers which are not in the cache.

package.json
lnist: yes please :)

lnist: I guess its fine :)

lnist (approved):

PR #1635, opened 2023-11-29
Feat/louis/separate clean summary from order

Louis-Amas:
https://app.asana.com/0/0/1205905634724379/f

### DIFF
--- a/src/market.ts
+++ b/src/market.ts
@@ -83,7 +83,8 @@ namespace Market {
   };
   export type OrderResult = {
     txReceipt: ethers.ContractReceipt;
-    summary: OrderSummary | CleanSummary;
+    summary?: OrderSummary;
+    cleanSummary?: CleanSummary;
     successes: Success[];
     tradeFailures: Failure[];
     posthookFailures: Failure[];
--- a/src/market.ts
+++ b/src/market.ts
@@ -81,7 +81,7 @@ namespace Market {
     bounty?: BigNumber;
     offersCleaned?: number;
   };
-  export type OrderResult = {
+  export type DirtyOrderResult = {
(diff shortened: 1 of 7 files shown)


test/integration/market.integration.test.ts
```diff
@@ -912,9 +912,9 @@ describe("Market integration tests suite", () => {
       fillVolume: 10,
     });
     const result = await buyPromises.result;
-    result.summary = result.summary as Market.OrderSummary;
+    result.summary = result.summary!;
+
     expect(result.tradeFailures).to.have.lengthOf(0);
-    expect(result.successes).to.have.lengthOf(1);
```
lnist, line 917: why removed?

src/market.ts
```diff
@@ -83,7 +83,8 @@ namespace Market {
   };
   export type OrderResult = {
     txReceipt: ethers.ContractReceipt;
-    summary: OrderSummary | CleanSummary;
```
lnist, line 86: I think it would make sense to have a completely different type than OrderResult for clean. CleanResult?
That would also avoid a bunch of ! in the tests

test/integration/market.integration.test.ts
Louis-Amas: Fixed

src/market.ts
lnist: What if you create a separate OrderResult and CleanResult which you return as the final result, but internally in tradeEventManagement you keep the ugly structure for now - could it then be a minimal change?

Louis-Amas: Yes we could do something like that easily

lnist (approved):

PR #1637, opened 2023-11-29
feat: make totalGave, totalGot, partialFill not undefined

Louis-Amas:
https://app.asana.com/0/0/1205905634724379/f


lnist (commented):
This one seems to have compile errors?

PR #1643, opened 2023-12-01
fix(market): Allow creation of non-existing market

lnist:


espendk (commented):
Small nit, other lgtm

espendk (approved):

PR #1648, opened 2023-12-04
feat: Market API simplification & fixes

espendk:
- `Market.Key` type introduced to identify markets (base, quote, tick spacing) in the API.
- Use of `Big(ish)` for `tickSpacing` in a few places has been replaced by `BigNumber(ish)` for consistency with the rest of the API.
- `Mangrove.openMarkets` no longer connects to all markets, but returns a list of `Market.Key`s, optionally with the relevant offer list configuration attached to each. This is identical to the previous `Mangrove.openMarketsData` which has been removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Next version
 
+- feat: `Market.Key` type introduced to identify markets (base, quote, tick spacing) in the API.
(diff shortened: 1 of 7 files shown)


dontrolle (approved):
lgtm

PR #1649, opened 2023-12-05
feat: Limit price and max tick improvements

espendk:
- Rename `price` -> `limitPrice` and `tick` -> `maxTick` where relevant.
- Rename the unused `boundary` parameter in `estimateVolume` to `limitPrice` and use it.


src/semibook.ts
```diff
@@ -164,16 +165,20 @@ namespace Semibook {
     Market.Offer[],
     LogSubscriber.Error
   >;
+
+  // Based on the OfferWrite event to ensure consistency and ease
+  // when mapping from raw to represensation
```
lnist, line 170: pregante?

lnist (approved):

PR #1651, opened 2023-12-05
feat: update take overhead

Louis-Amas:
Computed using data from:
https://github.com/mangrovedao/mangrove-strats/pull/155

### DIFF
--- a/src/constants/mangroveOrder.json
+++ b/src/constants/mangroveOrder.json
@@ -1,10 +1,10 @@
 {
-  "takeGasOverhead": 200000,
+  "takeGasOverhead": 330000,
   "restingOrderGasreq": 152000,
   "networks": {
     "maticmum": {
       "restingOrderGasreq": 152001,
-      "takeGasOverhead": 200001
+      "takeGasOverhead": 330000
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/constants/mangroveOrder.json
```diff
@@ -1,10 +1,10 @@
 {
   "takeGasOverhead": 200000,
-  "restingOrderGasreq": 152000,
+  "restingOrderGasreq": 330000,
```
lnist, line 3: restingOrderGasreq should not be changed.
It is takeGasOverhead that should be changed

Louis-Amas: Yep

test/unit/configuration.unit.test.ts
```diff
@@ -82,11 +82,11 @@ describe("Configuration unit tests suite", () => {
     );
     assert.equal(
       configuration.mangroveOrder.getTakeGasOverhead("local"),
-      200000,
+      330000,
     );
     assert.equal(
       configuration.mangroveOrder.getTakeGasOverhead("maticmum"),
-      200001,
+      330000,
```
lnist, line 89: add 1.
The value is intentionally different to test that configuration is correctly read

lnist (approved):

PR #1652, opened 2023-12-05
feat: lazy load olkeys mapping

Louis-Amas:


test/integration/mangrove.integration.test.ts
```diff
@@ -139,27 +139,4 @@ describe("Mangrove integration tests suite", function () {
       assert.equal(balance.toNumber(), amount);
     });
   });
-
-  describe("calculateOLKeyHash", () => {
```
lnist, line 143: why are you deleting this test?

src/mangrove.ts
```diff
@@ -366,8 +350,19 @@ class Mangrove {
     }
     return value;
   }
-  getOlKeyStruct(olKeyHash: string): OLKeyStruct | undefined {
-    return this.olKeyHashToOLKeyStructMap.get(olKeyHash);
+
+  async getOlKeyStruct(olKeyHash: string): Promise<OLKeyStruct | undefined> {
```
lnist, line 354: Additionally we should extend the getOlKeyHash to insert into the cache so we avoid unneeded roundtrips

PR #1658, opened 2023-12-06
chore(tick): Tick as number instead of BigNumber

lnist:


src/util/coreCalculations/Constants.ts
espendk, line 1: I think we should keep the mechanically translated core calculations files as is. And then map to/from in the code that interact with them.

espendk: Wrt. MAX_TICK and MIN_TICK which are referenced outside of Density and TickPriceHelper (the wrappers around core calculations): We can export them as `number` in TickPriceHelper instead. This way only imports need to be updated.

espendk (changes requested):
lgtm. Don't think we should change the core calculations code though, see the comments on Constants.ts

src/util/coreCalculations/Constants.ts
```diff
@@ -65,8 +65,8 @@ export const MIN_BIN = BigNumber.from("-1048576");
 export const MAX_BIN = BigNumber.from("1048575");
 
 /* The tick range is the largest such that the mantissa of `1.0001^MAX_TICK` fits on 128 bits (and thus can be multiplied by volumes). */
-export const MIN_TICK = BigNumber.from("-887272");
-export const MAX_TICK = BigNumber.from("887272");
+export const MIN_TICK = BigNumber.from(-887272);
+export const MAX_TICK = BigNumber.from(887272);
```
espendk, line 69:
```suggestion
export const MIN_TICK = BigNumber.from("-887272");
export const MAX_TICK = BigNumber.from("887272");
```

src/util/coreCalculations/TickLib.ts
```diff
@@ -117,7 +117,7 @@ export type Tick = int;
 // library TickLib {
 
   export function inRange(tick: Tick): boolean {
-    return tick >= MIN_TICK && tick <= MAX_TICK;
+    return tick.gte(MIN_TICK) && tick.lte(MAX_TICK);
```
espendk, line 120: good catch!

src/util/coreCalculations/Constants.ts
lnist: Very few places used them, so I decided to just do a toNumber

espendk (changes requested):
small nit on the Constants change revert ;-)

espendk (approved):

PR #1660, opened 2023-12-06
chore(tickSpacing): Thread tickSpacing throughout code

lnist:


src/util/tickPriceHelper.ts
```diff
@@ -9,20 +9,15 @@ import { MANTISSA_BITS, MIN_RATIO_EXP } from "./coreCalculations/Constants";
 
 class TickPriceHelper {
   ba: Market.BA;
-  market: {
-    base: { decimals: number };
-    quote: { decimals: number };
-  };
+  //FIXME use market.tickSpacing and TickLib.nearestBin to calculate prices and ticks with correct precision given the tick spacing - and extend tests
```
espendk, line 12: Is this intentionally left in?

src/market.ts
```diff
@@ -34,6 +34,12 @@ namespace Market {
     tickSpacing: number;
   };
 
+  export type KeyData = {
```
espendk, line 37: Since this type only contains part of a `Market.Key`'s data (eg token address and symbol are omitted), this name feels misleading. Don't get me wrong, I think introducing a type for this subset of data which is used in many places is a big improvement (though I'm still unhappy about the pattern, but that's another discussion 😅).

Could we call it something more specific?
Here are some ideas (not perfect, but less misleading):
1. `Market.KeyDecimalsAndTickSpacing`
2. `Market.TickPriceHelperData`
3. `Market.DecimalsAndTickSpacing`.

I think 3. is the better of those ideas.

espendk (commented):
lgtm except for the name of the `Market.KeyData` type, see the comment on market.ts.

src/market.ts
espendk: And we should probably add a comment that explains the purpose of that somewhat weird and technical type :-)

src/util/tickPriceHelper.ts
lnist: However, since we now have asana issue on it, I've removed it.

src/market.ts
lnist: Changed name, but also added yet another commit on top with a slightly different, and hopefully better, design.

test/unit/kandel/kandelDistribution.unit.test.ts
```diff
@@ -81,8 +82,9 @@ describe(`${KandelDistribution.prototype.constructor.name} unit tests suite`, ()
           asks: [{ gives: Big(0), tick: 1, index: 1 }],
         },
         {
-          base: { decimals: 4 },
-          quote: { decimals: 6 },
+          base: new TokenCalculations(4, 4),
+          quote: new TokenCalculations(6, 6),
+
```
espendk, line 87: Is GH messing with me or is this a misplaced empty line? I can't propose to delete it...

espendk (commented):
Big improvement in clarity. Not perfect, but much better :-)

Added some smaller comments/questions.

espendk (approved):

PR #1662, opened 2023-12-08
feat(tick): Coerce ticks to nearest exact tick wrt tickSpacing

lnist:


src/util/tickPriceHelper.ts
```diff
@@ -128,22 +139,51 @@ class TickPriceHelper {
...
+   */
+  public isTickExact(tick: number): boolean {
+    return this.nearestRepresentableTick(BigNumber.from(tick)).eq(tick);
+  }
+
+  /** Coerce a tick to its nearest bin
+   * @param tick tick to coerce
+   * @returns tick coerced to its nearest bin
+   */
+  private nearestRepresentableTick(tick: BigNumber): BigNumber {
```
espendk, line 169: Why `BigNumber` here when tick is number elsewhere?

espendk (commented):
Great first step. Uncontrolled coercion is probably not what we want, but at least now it is clear where coercion is happening.

Added a single question.

src/util/tickPriceHelper.ts
lnist: Because it is a private helper wrapping TickLib, thus reduces conversions to keep it as BigNumber.

src/util/tickPriceHelper.ts
```diff
@@ -128,22 +139,51 @@ class TickPriceHelper {
...
+    return TickLib.nearestBin(
+      BigNumber.from(tick),
```
espendk, line 171:
```suggestion
      tick,
```

src/util/tickPriceHelper.ts
espendk: Alright. See suggestion below

espendk (approved):

PR #1663, opened 2023-12-08
feat(Kandel): Coerce price-based parameters to exact ticks, and validate only exact ticks used for distribution

lnist:
Commits can be reviewed individually


src/kandel/geometricKandel/geometricKandelDistributionHelper.ts
```diff
@@ -104,8 +109,8 @@ class GeometricKandelDistributionHelper {
    * @returns The tick based parameters, @see TickDistributionParams
    */
   public getTickDistributionParams(
-    params: DistributionParams,
-  ): TickDistributionParams {
+    params: Omit<Omit<DistributionParams, "stepSize">, "generateFromMid">,
```
Louis-Amas, line 112: Why are you using double Omit ?
Can't you do `Omit<DistributionParams, "stepSize" | "generateFromMid">` ?

Louis-Amas (approved):

PR #1665, opened 2023-12-08
feat: Cache global and local configs

espendk:
Mangrove and Semibook configs are now cached on `connect` and (for Semibook) updated by events.

The methods to read configs are no longer async and naming has been made consistent: `Mangrove.config()`, `Market.config()`, and `Semibook.config()`.

Caching the configs limits the RPC requests sent from frequently called methods, in particular `Semibook.simulateMarketOrder` which is used to eg estimate trading volumes and gas.

As part of this change, the following smaller changes have been made:
- Two missing global config fields have been added: `maxRecursionDepth` and `maxGasreqForFailingOffers`.
- All order book events now carry the relevant data for the event. And events that are not related to offers do not carry offer data.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205923576109048

### DIFF
--- a/src/semibook.ts
+++ b/src/semibook.ts
@@ -173,6 +173,7 @@ namespace Semibook {
    * - isComplete                                                         =>  all offers in the offer list are in the cache
    */
   export type State = {
+    localConfig: Mangrove.LocalConfig; // local config for the offer list
     offerCache: Map<number, Market.Offer>; // offer ID -> Offer
     binCache: Map<number, Semibook.Bin>; // tick -> Bin
     bestBinInCache: Semibook.Bin | undefined; // best/first bin in the offer list iff #binCache is non-empty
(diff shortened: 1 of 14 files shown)


src/semibook.ts
```diff
@@ -291,8 +293,17 @@ class Semibook
         ? undefined
         : clonedBinCache.get(state.worstBinInCache.tick);
 
+    const clonedLocalConfig = {
+      ...state.localConfig,
+      density: new Density(
```
lnist, line 298: Add .clone to Density?

espendk: Addressed in 66d3b47f99998d0a22acd838ada3ebce451cd968

lnist (approved):

PR #1666, opened 2023-12-09
feat: Implement MGV v2 simulateMarketOrder

espendk:
The market order simulation used to estimate volumes and gas has been updated to match Mangrove v2's market order logic.

As part of this, `Market.estimateVolume` now also estimates fees and returns it in a new `estimatedFee` field. The existing `estimatedVolume` field is exclusive of fees and thus represents the true amount the taker can expect to receive/pay.

### DIFF
--- a/src/semibook.ts
+++ b/src/semibook.ts
@@ -478,7 +478,7 @@ class Semibook
       ? this.tickPriceHelper.tickFromPrice(params.limitPrice)
       : MAX_TICK.toNumber();
 
-    const { totalGot, totalGave, feePaid, maxTickMatched } =
+    const { totalGot, totalGave, feePaid, fillVolume, maxTickMatched } =
       await this.simulateMarketOrder(maxTick, initialVolume, buying);
 
     const estimatedVolume = buying ? totalGave : totalGot;
@@ -486,7 +486,7 @@ class Semibook
     return {
       maxTickMatched,
       estimatedVolume: estimatedVolume,
-      remainingFillVolume: initialVolume.sub(estimatedVolume),
+      remainingFillVolume: fillVolume,
       estimatedFee: feePaid,
     };
   }
@@ -506,6 +506,7 @@ class Semibook
     totalGot: Big;
     totalGave: Big;
     feePaid: Big;
+    fillVolume: Big;
     maxTickMatched?: number;
     gas: BigNumber;
   }> {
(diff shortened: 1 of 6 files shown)


src/semibook.ts
```diff
@@ -481,111 +481,185 @@ class Semibook
...
-      await this.simulateMarketOrder(maxTick, initialGives, buying);
+    const {
+      maxTickMatched,
+      remainingFillVolume,
+      totalGot,
+      totalGave,
+      feePaid,
+    } = await this.simulateMarketOrder(maxTick, initialGives, buying);
 
     const estimatedVolume = buying ? totalGave : totalGot;
```
lnist, line 492: seems like estimatedVolume could be set on the return of this.simulateMarketOrder to avoid all this unpacking and packing of values?

test/integration/market.integration.test.ts
```diff
@@ -1188,39 +1155,137 @@ describe("Market integration tests suite", () => {
...
+        quote: "TokenB",
+        tickSpacing: 1,
+      });
+
+      const done = new Deferred();
+      let volumeEstimate: Market.VolumeEstimate | undefined;
+      const baseVolume = 0.5;
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      market.subscribe(async (evt) => {
+        if (market.getBook().asks.size() === 2) {
```
lnist, line 1185: why is this deferring needed? waitForbLock should be enough?

src/semibook.ts
espendk: I prefer to keep `simulateMarketOrder` close to the Solidity code and then add functionality/mapping in `estimateVolume`.

But I agree it was a bit messy, so I've simplified a bit here: 2d530a65e048ac992d7f6c53640ed60a7d17fa15

test/integration/market.integration.test.ts
espendk: It isn't, I just didn't clean up that old code :-)

Addressed here: 39a9b320c34a577bf8f17701cd7f4b22ec7de640

src/semibook.ts
espendk: made a mistake, so here's a fix commit: 9b7dd9d1d051fd42c914b743c511cb0a86abfa4b

lnist (approved):

PR #1669, opened 2023-12-11
feat: Add `retractRestingOrder` method to Market

espendk:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205905634724382


lnist (approved):
I would usggest you use the s

PR #1670, opened 2023-12-11
feat(MangroveOrder): Calculate default provision for MangroveOrder

lnist:


src/util/trade.ts
```diff
@@ -626,16 +629,14 @@ class Trade {
           fillOrKill: fillOrKill,
           tick: maxTick,
           fillVolume: fillVolume,
-          fillWants: orderType === "buy",
-          restingOrder: postRestingOrder,
+          fillWants: fillWants,
+          restingOrder: restingOrderParams ? true : false,
```
espendk, line 633:
```suggestion
          restingOrder: !!restingOrderParams,
```

src/util/trade.ts
```diff
@@ -711,18 +712,56 @@ class Trade {
...
+  public async getRestingOrderParams(
+    params: Market.RestingOrderParams,
+    market: Market,
+    ba: Market.BA,
+  ): Promise<{
+    provision: Big.BigSource;
+    restingOrderGasreq: number;
+    gaspriceFactor: number;
+    restingOrderBa: string;
+  }> {
```
espendk, line 730: ❤️

test/integration/restingOrder.integration.test.ts
```diff
@@ -130,38 +130,72 @@ describe("RestingOrder", () => {
...
+          },
+        });
+        const tx = await waitForTransaction(buyPromises.response);
+        await mgvTestUtil.waitForBlock(mgv, tx.blockNumber);
+
+        const orderResult = await buyPromises.result;
+        orderResult.summary = orderResult.summary as Market.OrderSummary;
+        assert(
+          // 2.5% fee configured in mochaHooks.js
+          orderResult.summary.totalGot!.eq(10 * 0.975),
```
espendk, line 160: You could read the fee from `orderLP.market`

espendk: Same further down. I'm afraid of how maintainable this is with constants scattered instead of defined and calculated once..

espendk (commented):
nice solution 👌

Added some comments

test/integration/restingOrder.integration.test.ts
lnist: It seems good enough, we have not changed this recently, and I did not add it, so test cleanup is another PR :D

espendk (approved):

PR #1675, opened 2023-12-12
chore(conversion): Use tickPriceHelper when available for outbound/inbound raw conversions

lnist:


src/util/tickPriceHelper.ts
```diff
@@ -5,10 +5,13 @@ import Market from "../market";
 import Big from "big.js";
 import { Bigish } from "../types";
 import { MANTISSA_BITS, MIN_RATIO_EXP } from "./coreCalculations/Constants";
+import { TokenCalculations } from "../token";
 
 class TickPriceHelper {
   ba: Market.BA;
   market: Market.KeyResolvedForCalculation;
+  outbound: TokenCalculations;
```
dontrolle, line 13: These four fields should really be `readonly`, right? 

The need becomes more pronounced when we start caching fields calculated from other fields.

src/util/tickPriceHelper.ts
```diff
@@ -18,6 +21,9 @@ class TickPriceHelper {
   constructor(ba: Market.BA, market: Market.KeyResolvedForCalculation) {
     this.ba = ba;
     this.market = market;
+
+    this.outbound = this.ba === "bids" ? this.market.quote : this.market.base;
```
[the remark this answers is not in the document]
dontrolle: One could also calculating inbound and outbound each time they're needed below. I think it's ok to cache them - but probably a good idea to add readonly'ness of the fields.

dontrolle (changes requested):
Partial review of first changeset.

src/util/tickPriceHelper.ts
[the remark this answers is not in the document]
dontrolle: Update: With revised changeset - see however comment on Semibook that it looks weird to get inbound/outbound from TickPriceHelper - and it may, e.g., make more sense to cache inbound/outbound token on Semibook (if any caching is to be done).

src/kandel/coreKandelInstance.ts
```diff
@@ -310,12 +310,16 @@ class CoreKandelInstance {
     const rawDistribution: KandelTypes.DirectWithBidsAndAsksDistribution.DistributionStruct =
       {
         bids: distribution.bids.map((o) => ({
-          gives: this.market.quote.toUnits(o.gives),
+          gives: this.distributionHelper.bidTickPriceHelper.rawOutbound(
```
dontrolle, line 313: I understand that we've added some functionality concerning price conversions to `TickPriceHelper`, But I'm wondering whether we're taking it a bit too far here. 

We've also decided to collect some functionality concerning conversion of "raw" values to "non-raw" values on `Token` (and `TokenCalculation`). I'm not sure adding the wrapper function and using it like here on `TickPriceHelper` actually improves readability? 

At the least I need a bit of convincing.

dontrolle: In addition - might we consider improving the naming of the `toUnits` function, instead? Say to `toRawUnits` - or something like that?

src/util/tickPriceHelper.ts
```diff
@@ -114,15 +116,41 @@ class TickPriceHelper {
...
     const rawOutbound = (
       roundUp ? TickLib.outboundFromInboundUp : TickLib.outboundFromInbound
     )(bin, rawInbound);
-    return (
-      this.ba === "bids" ? this.market.quote : this.market.base
-    ).fromUnits(rawOutbound);
+    return this.outboundFromRaw(rawOutbound);
+  }
+
+  /** Converts outbound token amount from public amount to internal contract amount.  */
```
dontrolle, line 126: See comments for `src/kandel/coreKandelInstance.ts`

lnist:
Closing in favor of doing a well-designed overhaul later.

PR #1676, opened 2023-12-13
fix(semibook): Use outbound/inbound in place of manual derivation

lnist:


src/util/tickPriceHelper.ts
```diff
@@ -20,6 +20,24 @@ class TickPriceHelper {
     this.market = market;
   }
 
+  /** Gets the outbound token */
```
dontrolle, line 23: In my view, I'd like that these two functions be convenience functions used privately by the `TickPriceHelper` (as they also are used in this PR); basically so we don't spread out usage of `myTickPriceHelper.getOutBound()` various places in the code. 

I look forward to a refactoring, where we find a better way to locate some of this functionality not an a helper-class like, `TickPriceHelper`, but more naturally on `Semibook`, or on a "TickPrice"-class. For right now, I'd look not to open up for more confusion.

dontrolle (changes requested):

dontrolle (approved):

PR #1698, opened 2024-01-04
feat(market): Add updateRestingOrder function

espendk:


src/market.ts
```diff
@@ -982,6 +1004,21 @@ class Market {
     return v ?? BigNumber.from(0);
   }
 
+  /** Update a resting order posted by MangroveOrder.
+   *
+   * @param ba whether the offer is a bid or ask
+   * @param params update parameters - see {@link Market.UpdateRestingOrderParams}
```
lnist, line 1010: Is see @link better than @see ?

lnist: Sorry @link and @see for mentioning you :)

src/util/trade.ts
```diff
@@ -308,8 +308,129 @@ class Trade {
...
+   * Gets parameters to send to function `market.mgv.orderContract.updateOffer`.
+   *
+   * @param params update parameters - see {@link Market.UpdateRestingOrderParams}
+   * @param market the market to retract the order on
+   * @param ba whether the offer is a bid or ask
+   * @returns a promise that resolves to the raw parameters to send to the MangroveOrder contract
+   *
+   * @remarks
+   *
+   * @see {@link Market.CleanParams} for a more thorough description of cleaning parameters.
```
lnist, line 374: Why a reference to CleanParams? copy-pasta?

src/util/trade.ts
```diff
@@ -308,8 +308,129 @@ class Trade {
...
+   */
+  async getRawUpdateRestingOrderParams(
```
lnist, line 376: This looks a lot like what we have in other params functions - can't we de-duplicate everything around tick/price/volume?

espendk: I attempted this initially as there's clearly an overlap, but the parameter options differ significantly, so finding the common ground and refactoring is a non-trivial task and not something we have time for at the moment.

src/market.ts
espendk: According to this site, `see` is a block tag for references whereas `link` is for linking: https://tsdoc.org/pages/tags/see/

So I think the usage here is correct.

lnist (approved):

PR #1701, opened 2024-01-05
feat(tickSpacing): Round up, down, or to nearest depending on scenario

lnist:
Review commit-by-commit, see header.

`feat(tickSpacing): Add rounding up/down` contains the bulk of the rounding changes and some test changes (and a ninja-rename), but subsequent commits add more tests.

Semibook changes are not tested as we have no unit tests for the changed functions and we have no market with tickSpacing>1

### DIFF
--- a/src/util/tickPriceHelper.ts
+++ b/src/util/tickPriceHelper.ts
@@ -6,7 +6,7 @@ import Big from "big.js";
 import { Bigish } from "../types";
 import { MANTISSA_BITS, MIN_RATIO_EXP } from "./coreCalculations/Constants";
 
-/** roundDown rounds toward zero. roundUp rounds away from zero, and nearest tries to round to the nearest being towards or away from zero. */
+/** roundDown rounds down to a representable value, roundUp rounds up to a representable value, and nearest rounds to the nearest representable value. */
 export type RoundingMode = "nearest" | "roundDown" | "roundUp";
 
 class TickPriceHelper {
@@ -42,9 +42,22 @@ class TickPriceHelper {
   }
 
   /**
+   * Inverses the rounding mode for bids.
+   */
+  #inverseRoundingModeForBids(roundingMode: RoundingMode) {
+    return this.ba === "asks"
+      ? roundingMode
+      : roundingMode === "roundDown"
+        ? "roundUp"
+        : roundingMode === "roundUp"
+          ? "roundDown"
+          : roundingMode;
+  }
+
+  /**
    * Calculates the price at a given raw offer list tick.
    * @param tick tick to calculate price for (is coerced to nearest bin)
-   * @param roundingMode the rounding mode for coercing tick to a representable tick. See {@link RoundingMode}. Up is to a higher price, down is to a lower price.
+   * @param roundingMode the rounding mode for coercing tick to a representable tick. See {@link RoundingMode}. roundDown is to a lower price, roundUp is to a higher price.
    * @returns price at tick (not to be confused with offer list ratio).
    */
   priceFromTick(tick: number, roundingMode: RoundingMode): Big {
@@ -52,7 +65,10 @@ class TickPriceHelper {
     const dp = Big.DP;
     Big.DP = 300;
 
-    const offerListRatioFromTick = this.rawRatioFromTick(tick, roundingMode);
+    const offerListRatioFromTick = this.rawRatioFromTick(
+      tick,
+      this.#inverseRoundingModeForBids(roundingMode),
+    );
     // For scaling the price to the correct decimals since the ratio is for raw values.
     const decimalsScaling = Big(10).pow(
       this.market.base.decimals - this.market.quote.decimals,
@@ -72,7 +88,7 @@ class TickPriceHelper {
   /**
    * Calculates the raw offer list tick (coerced to nearest bin) at a given order book price (not to be confused with offer list ratio).
    * @param price price to calculate tick for
-   * @param roundingMode See {@link RoundingMode} Lower is to a lower tick, higher is to a higher tick.
+   * @param roundingMode See {@link RoundingMode} roundDown is to a lower tick, roundUp is to a higher tick.
    * @returns raw offer list tick (coerced to nearest bin) for price
    */
   tickFromPrice(price: Bigish, roundingMode: RoundingMode): number {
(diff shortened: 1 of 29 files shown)


src/util/tickPriceHelper.ts
```diff
@@ -6,6 +6,9 @@ import Big from "big.js";
 import { Bigish } from "../types";
 import { MANTISSA_BITS, MIN_RATIO_EXP } from "./coreCalculations/Constants";
 
+/** roundDown rounds toward zero. roundUp rounds away from zero, and nearest tries to round to the nearest being towards or away from zero. Default is nearest. */
```
espendk: > nearest *tries* to round to the nearest

Can it fail?

> to the nearest being towards or away from zero.

I cannot quite parse/understand this, can you explain?

src/util/tickPriceHelper.ts
```diff
@@ -89,49 +95,83 @@ class TickPriceHelper {
...
    * @returns the price coerced to nearest representable tick */
-  coercePrice(price: Bigish): Big {
-    const tick = this.tickFromPrice(price);
-    return this.priceFromTick(tick);
+  coercePrice(price: Bigish, roundingMode: RoundingMode): Big {
+    let tick = this.tickFromPrice(price, roundingMode);
+    let coercedPrice = this.priceFromTick(tick, roundingMode);
+    if (roundingMode === "roundDown") {
+      while (coercedPrice.gt(price)) {
+        tick -= this.market.tickSpacing;
```
espendk, line 110: For bids, doesn't the price get larger when the tick becomes smaller? Since for bids, price = 1/ratio.

So for bids, I think the tick should be increased here.

Vice versa in the code for rounding down below.

src/util/tickPriceHelper.ts
```diff
@@ -149,44 +189,104 @@ class TickPriceHelper {
...
+      }
+    } else if (roundingMode === "roundUp") {
+      while (
+        this.inboundFromOutbound(binnedTick, outboundVolume, roundingMode).lt(
+          inboundVolume,
+        )
+      ) {
+        binnedTick += this.market.tickSpacing;
+      }
+    }
```
espendk, line 219: I think the comment on `coercePrice` doesn't apply here, since this code does only deals with inbound/outbound and thus doesn't have to handle th asymmetry that prices introduces. Is that right?

src/util/tickPriceHelper.ts
```diff
@@ -149,44 +189,104 @@ class TickPriceHelper {
...
+   * @param roundingMode the rounding mode for coercing tick to a representable tick. @see RoundingMode, default is nearest.
    * @returns tick coerced to its nearest bin
    */
-  private nearestRepresentableTick(tick: BigNumber): BigNumber {
-    return TickLib.nearestBin(
+  private nearestRepresentableTick(
+    tick: BigNumber,
+    roundingMode: RoundingMode,
+  ): BigNumber {
+    // Note: nearestBin is a misnomer - it rounds
```
espendk, line 258: I don't quite understand this comment, can you elaborate?

src/util/trade.ts
```diff
@@ -44,7 +44,7 @@ class Trade {
         slippage,
         "buy",
       );
-      maxTick = tickPriceHelper.tickFromPrice(priceWithSlippage);
+      maxTick = tickPriceHelper.tickFromPrice(priceWithSlippage, "roundDown");
```
espendk, line 47: Maybe add the same comment here re. rounding as in the other places?

>// round down to not exceed the price

src/util/trade.ts
```diff
@@ -76,9 +84,11 @@ class Trade {
       );
       fillWants = params.fillWants ?? true;
       fillVolume = fillWants ? Big(params.wants) : givesWithSlippage;
+      // round down to not exceed price expectations
       maxTick = tickPriceHelper.tickFromVolumes(
         givesWithSlippage,
         params.wants,
+        "roundDown",
```
espendk, line 91: Does rounding down the tick correspond to rounding down the price in this context? Ie. is `givesWithSlippage` in base and `wants` in quote?

src/util/trade.ts
```diff
@@ -157,9 +173,11 @@ class Trade {
       );
       fillWants = params.fillWants ?? false;
       fillVolume = fillWants ? wantsWithSlippage : Big(params.gives);
+      // round down to not exceed the price expectations
       maxTick = tickPriceHelper.tickFromVolumes(
         params.gives,
         wantsWithSlippage,
+        "roundDown",
```
espendk, line 180: Same question as above: Does rounding down the tick correspond to rounding down the price in this context? Ie. is `gives` in base and `wantsWithSlippage` in quote?

src/util/trade.ts
```diff
@@ -397,24 +415,31 @@ class Trade {
     }
 
     if ("price" in params && params.price !== undefined) {
-      tick = tickPriceHelper.tickFromPrice(Big(params.price));
+      // round up to ensure we get at least the price we want for the offer.
+      tick = tickPriceHelper.tickFromPrice(Big(params.price), "roundUp");
```
espendk, line 420: If this is a bid, I think we should be rounding down so we don't pay more than what we wanted to pay.

src/util/trade.ts
espendk: I wonder whether we actually always want to round down when calling `tickFromPrice` or if it depends on whether we're in a bids or asks context?

espendk (commented):
Very nice 👌

I think there may some issues due to the asymmetry of asks and bids, but I'm not entirely sure of the extent. Maybe let's discuss on a call?

src/util/trade.ts
lnist: as discussed offline it is correct as is.

src/util/trade.ts
lnist: same as above.

src/util/trade.ts
lnist: We are the maker, so we need to round up

src/util/trade.ts
lnist: Discussed offline, and agree current code is correct.

src/util/tickPriceHelper.ts
lnist: Agree, fixed in 467155c2168f261e91976414be8d0af7d050daa4

src/util/tickPriceHelper.ts
lnist: Yes :)

espendk (approved):

PR #1713, opened 2024-01-10
Update to new Mangrove Order

maxencerb:
TODO list:

- [x] Match the new function signatures of Mangrove order
- [x] Add deploy aave logic on deploy script
- [x] Add resting order gas req dependant on logic
- [x] Add test for aave logic
- [x] Add setRouteLogic for an offer
- [x] Provide a set of strats out of the box


src/market.ts
```diff
@@ -1440,6 +1486,61 @@ class Market {
...
+        olKeyHash,
+        token: params.token,
+        offerId: params.offerId,
+        fundOwner: ethers.constants.AddressZero, // is not useful for this function
+      },
+      params.logic,
+      overrides,
+    );
+    const res = await tx.wait();
+    return res;
```
espendk, line 1542: For consistency, maybe this function should return a `Market.Transaction<T>` like the other market functions that send transactions?

src/util/trade.ts
```diff
@@ -299,7 +304,7 @@ class Trade {
     if (offer === undefined) {
       throw Error(`No offer in market with id ${id}.`);
     }
-    if (offer.maker !== market.mgv.orderContract.address) {
+    if (offer.maker !== (market.mgv.orderContract as any).address) {
```
espendk, line 307: How come `as any` is needed here? `Mangrove.orderContract` does not seem to be changed by this PR?

src/util/trade.ts
```diff
@@ -663,6 +681,31 @@ class Trade {
...
+  } {
+    switch (orderType) {
+      case MangroveOrderType.FOK:
+        return { fok: true, restingOrder: false };
+      case MangroveOrderType.GTC:
+        return { fok: false, restingOrder: true };
+      default:
+        return { fok: false, restingOrder: false };
+    }
+  }
```
espendk, line 708:
```suggestion
  }

```

src/util/trade.ts
```diff
@@ -805,34 +856,73 @@ class Trade {
...
     gaspriceFactor: number;
     restingOrderBa: string;
   }> {
+    let restingOrderGasreq =
+      params.restingOrderGasreq ??
+      configuration.mangroveOrder.getRestingOrderGasreq(
+        market.mgv.network.name,
+      );
+    if (!params.restingOrderGasreq && logics) {
+      const takerWants = market.mgv.getLogicByAddress(logics.takerWantsLogic);
```
espendk, line 882: nit: `takerWants` and `takerGives` are usually volumes in Mangrove code. Here it seems they are logics, so maybe the naming feels misleading.

src/util/trade.ts
```diff
@@ -805,34 +856,73 @@ class Trade {
...
-      configuration.mangroveOrder.getRestingOrderGasreq(
-        market.mgv.network.name,
-      );
 
     const restingOrderBa = ba === "asks" ? "bids" : "asks";
     let provision = params.provision;
     if (!provision) {
       const mangroveOrder = market.mgv.offerLogic(
-        market.mgv.orderContract.address,
+        (market.mgv.orderContract as any).address,
```
espendk, line 925: Again: Why `as any` here?

test/mocha/config/market-tests.json
espendk, line 1: Since this only runs restingOrder integration tests, the name should reflect that.

And maybe there doesn't need to be a separate configuration for the resting order tests? Would Mocha's `-g` option be sufficient? Or `.only`?

src/logics/AbstractRoutingLogic.ts
```diff
@@ -0,0 +1,96 @@
...
+    tokenAddress: string,
+  ): Promise<string>;
+
+  /**
+   * @desc Returns the overlying address for a token.
+   * * It will first check the cache, and if it is not there, it will query the network.
+   * @param tokenAddress the address of the token
+   * @returns The overlying address.
+   */
+  async overlyingAddress(tokenAddress: string): Promise<string> {
```
espendk, line 66: The SDK rarely passes raw token addresses around. Perhaps a `Token` parameter or token ID would be more appropriate here?

Same for the other functions that take a `tokenAddress` parameter.

espendk (commented):
Very nice!

Added some smaller comments.

package.json
```diff
@@ -18,6 +18,7 @@
     "clean": "rimraf dist",
     "test": "npm-run-all --parallel test:unit test:integration",
     "test:integration": "cross-env NODE_ENV=test nyc --reporter=lcov mocha --parallel --config test/mocha/config/integration-tests.json --exit",
+    "test:market": "cross-env NODE_ENV=test nyc --reporter=lcov mocha --parallel --config test/mocha/config/market-tests.json --exit",
```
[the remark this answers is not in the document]
maxencerb: This will be removed anyway, I just didn't knew there was a way to filter specific tests to run

src/logics/AbstractRoutingLogic.ts
```diff
@@ -0,0 +1,96 @@
...
+    return res;
+  }
+
+  /**
+   * @desc Returns whether or not the logic can be used for a token.
+   * @param tokenAddress the address of the token
+   * @returns Whether or not the logic can be used for the token.
+   */
+  async canUseLogicFor(tokenAddress: string): Promise<boolean> {
+    const address = await this.overlyingAddress(tokenAddress);
```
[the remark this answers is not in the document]
maxencerb: It now returns 0 if nothing is to be approved.
At the moment liquidity sharing is not thought about in the UI.
A later addition could be done to add this possibility to SDK and UI

src/mangrove.ts
```diff
@@ -138,6 +140,9 @@ class Mangrove {
 
   eventEmitter: EventEmitter;
   _config: Mangrove.GlobalConfig; // TODO: This should be made reorg resistant
+  logics: {
+    aave: SimpleAaveLogic;
```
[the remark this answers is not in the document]
maxencerb: Added `SimpleLogic` which now defaults

espendk (approved):
Nice improvements 👌

As you're off, I've taken the liberty of pushing some very minor adjustments, so we can get this merged and published.

DanielRX (approved):

PR #1719, opened 2024-01-11
feat(market): Add spread to market and getBest to semibook

lnist:


src/market.ts
```diff
@@ -778,6 +779,42 @@ class Market {
...
+    const bestAsk = await asks.getBest();
+    const bestBid = await bids.getBest();
+
+    return Market.spread(this, bestAsk, bestBid);
+  }
+
+  /**
+   * Gets the absolute, relative, and tick spread between a bid and an ask on the market.
+   */
+  public static spread(
```
espendk, line 797: why `public` here, but not in the previous method?
AFAIR the convention in `Market` is to omit the access modifier for `public` functions.

src/market.ts
```diff
@@ -778,6 +779,42 @@ class Market {
...
+    }
+    const lowestAskPrice = Big(bestAsk.price);
+    const highestBidPrice = Big(bestBid.price);
+    const absoluteSpread = lowestAskPrice.sub(highestBidPrice);
+    const tickSpread = bestAsk.tick + bestBid.tick;
+    // Intentionally using raw ratio as we do not want decimals scaling
+    // Rounding is irrelevant as ticks already respects tick spacing
+    const relativeSpread = new TickPriceHelper("asks", market)
+      .rawRatioFromTick(tickSpread, "roundUp")
+      .sub(1);
```
espendk, line 813: Hmm, does this correspond to $1.0001^{tickSpread} - 1$?
Maybe writing that calculation would be clearer?

espendk (commented):
nice. Added a few questions

src/market.ts
espendk: As discussed out-of-band, let's keep it as is

espendk (approved):

PR #1730, opened 2024-01-15
feat(doc): Update type doc to avoid warnings

lnist:


src/index.ts
espendk, line 1: This file has become quite large and I'm guessing that most of the exported types are irrelevant for clients of mangrove.js? But as I understand it, this is the only way you've found to ensure all the types are included in the docs?

How difficult do you think this will be to maintain?
Will we eg get an error whenever a new type is introduced, but not included here?

typedoc.json
espendk, line 1: There are a lot of things listed in this file and it's not entirely clear to me, why they are listed?
How should this be maintained?

src/index.ts
lnist: See answer to other comment.

typedoc.json
lnist: `yarn doc` will report errors if type is used by another, exported type. Then either add it to the intentionally not exported or the index.ts.

`yarn doc` is run by CI but not by `yarn build` so there currently is the downside that developers will get blindsided by an error in github actions which they have to run `yarn doc` locally to discover.

lnist: Oh, and to me, exporting a type is writing export { type } and then using the IDE to import it - so updating the imports in the file is not an issue to me :)

espendk: Alright. I think this is a big step towards a better setup, so let's merge it. We can always improve it later if we find a way to do it

espendk (approved):

PR #1745, opened 2024-01-18
feat: Add integration test of tickSpacing>1

lnist:
Also adds coercion of tick arguments, removed parallel test run, and fixes updateOffer on LiquidityProvider for an EOA.


src/util/tickPriceHelper.ts
```diff
@@ -246,6 +249,21 @@ class TickPriceHelper {
     return binnedTick;
   }
 
+  /** Calculates the tick offset closely corresponding to the given raw ratio.
+   * @param priceRatio the price ratio.
```
espendk, line 253: What is a "price ratio"?

src/util/trade.ts
```diff
@@ -62,26 +62,16 @@ class Trade {
...
+      maxTick = params.maxTick;
       if (slippage > 0) {
-        // round down to not exceed the price when buying, and up to get at least price for when selling
-        const limitPrice = tickPriceHelper.priceFromTick(
-          params.maxTick,
-          bs === "buy" ? "roundDown" : "roundUp",
+        // offset the tick by getting a non-rounded offset
+        maxTick += tickPriceHelper.tickOffsetFromRawRatio(
+          Big(1).sub(Big(100 + bs === "buy" ? slippage : -slippage).div(100)),
+          "none",
```
espendk, line 70: If this is the only use for `tickOffsetFromRawRatio`, then maybe we could replace it with a method that takes a tick and the percentage that should be added to it? Or something like that.

Right now, the method itself and the call site are somewhat difficult to understand. Would be good if the complexity was only in one place :-)

espendk (commented):
nice!

I'm unsure of some of the terminology/naming, I think it could be clearer. Let's talk about it.

src/util/tickPriceHelper.ts
```diff
@@ -208,17 +208,20 @@ class TickPriceHelper {
...
     outboundVolume: Bigish,
-    roundingMode: RoundingMode,
+    roundingMode: RoundingMode | "none",
   ): number {
     const rawInbound = this.#getInbound().toUnits(inboundVolume);
     const rawOutbound = this.#getOutbound().toUnits(outboundVolume);
     const tick = TickLib.tickFromVolumes(rawInbound, rawOutbound);
+    if (roundingMode === "none") {
+      return tick.toNumber();
+    }
```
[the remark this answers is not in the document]
lnist: Removed the option.

src/util/tickPriceHelper.ts
lnist: Concept from kandel - removed from here as function is deleted.

src/util/trade.ts
lnist: Replaced by usage of tickFromRawRatio and simplified expression.

src/util/tickPriceHelper.ts
```diff
@@ -246,6 +249,21 @@ class TickPriceHelper {
     return binnedTick;
   }
 
+  /** Calculates the tick offset closely corresponding to the given raw ratio.
+   * @param priceRatio the price ratio.
+   * @param roundingMode See {@link RoundingMode}. roundDown is to a lower tick, roundUp is to a higher tick. none is no rounding.
+   * @returns The tick offset corresponding to the price ratio.
+   */
+  tickOffsetFromRawRatio(priceRatio: Big, roundingMode: RoundingMode | "none") {
```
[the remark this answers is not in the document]
lnist: It is meant to mean a difference between two ticks which yields the equivalent ratio difference between their ratios. But function is deleted.

espendk (approved):
Nice improvements 👌

PR #1761, opened 2024-01-31
feat!: Rename Market.close to disconnect

espendk:
Rename `Market.close()` to `Market.disconnect()` to more clearly signal that it's dual to `Market.connect()` and does not close the market on Mangrove.


dontrolle (approved):
lgtm

PR #1764, opened 2024-02-06
Amplified Orders Support

DanielRX:


src/amplifier/mangroveAmplifier.ts
```diff
@@ -0,0 +1,219 @@
...
+  outboundLogic: z.instanceof(AbstractRoutingLogic),
+  expiryDate: liberalBigInt,
+  inboundTokens: z.array(inboundTokenSchema),
+});
+
+const updateBundleParams = z.object({
+  bundleId: liberalBigInt,
+  outboundToken: z
+    .string()
+    .refine((v) => ethers.utils.isAddress(v), "Invalid EVM Address"),
```
maxencerb, line 31: use `evmAddress` schema

src/amplifier/mangroveAmplifier.ts
```diff
@@ -0,0 +1,219 @@
...
+  outboundVolume: liberalBigInt,
+  updateExpiry: z.boolean(),
+  expiryDate: liberalBigInt,
```
maxencerb, line 34: Maybe we could check that it's zero or higher than current timestamp (timestamp on blockchain). Just a thought

maxencerb (commented):
There is something missing. Each offer from the bundle can be updated independently. However, we do not want the user to update the offer price. We want to allow modifying some things though.

One thing would be to modify the tick and the gasreq via `updateOffer` but not gives. We also want them to modify the logic that is used with router.setRouteLogic.

src/amplifier/mangroveAmplifier.ts
DanielRX: I considered this, but it would require zod to understand time on blockchain (which might differ per chain, so it's a bit hard to do cleanly

src/amplifier/mangroveAmplifier.ts
DanielRX: Done

src/amplifier/mangroveAmplifier.ts
maxencerb: Yeah that's what I figured. This would have to be done in the UI though because this will end up in failing to publish the bundle

maxencerb (approved):

maxencerb (approved):

PR #1798, opened 2024-02-26
fix amplifier export

maxencerb:


src/index.ts
```diff
@@ -184,3 +184,5 @@ export type {
   KandelNetworkConfiguration,
   Configuration,
 };
+
+export * from "./amplifier/mangroveAmplifier";
```
espendk, line 188: To be consistent with the rest of the file, there should be a section header, eg:

```suggestion
// Mangrove Amplifier
export * from "./amplifier/mangroveAmplifier";
```

espendk: Ideally, the import/export pattern would also be consistent within this file. Not sure how big a task that would be though.

espendk (dismissed):

espendk (approved):

espendk (approved):

PR #1851, opened 2024-03-11
return add bundle receipt

anasbelbaz:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next version
 
+# 2.0.5-40
```
Louis-Amas, line 3: Can you remove this ?

Louis-Amas (changes requested):

PR #1854, opened 2024-03-12
fix MemoryLimitOOG issue

espendk:


espendk:
Issue is caused by the latest forge nightly release (`forge 0.2.0 (7545c7a 2024-03-12T00:17:55.654014000Z)`). Others are reporting the same issue in the Foundry Support Telegram channel.

So closing this and hope that it'll be fixed in the next nightly release 🤞

PR #1856, opened 2024-03-13
build(deps): Update mangrove dependencies to latest versions

espendk:


espendk:
The test failures are due to a known bug in the latest Foundry releases, so merging this anyway.

